PR #4, opened 2018-10-12
Remove unnecessary dependencies

### LINKED ISSUE #3: Plugin is not compiling
Hi,

I just tried to use yours plugin, but unfortunately I'm getting error every time I'm try to run a project with it.

```
* Where:
Script '/Users/piotrw/Library/flutter/packages/flutter_tools/gradle/flutter.gradle' line: 210

* What went wrong:
A problem occurred configuring project ':parse_server_sdk'.
> Could not find method debugProvided() for arguments [file collection] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

I'm not sure, but from what I see in workspace with my IDE, and when I download whole archive from https://pub.dartlang.org/packages/parse_server_sdk#-versions-tab- there is lack of /android and /ios directories for this plugin. And You have those directories in your project on github. Maybe this is the reason of this error?

Could You take a look on that? Thanks in advance, and greetings!

PiotrWpl:
Fixes #3


phillwiggins:
What's your thoughts on this?

So, my understanding and reasoning behind this was that to have this library available via DartPub, this was required?

PiotrWpl:
When I installed yours plugin from DartPub it didn't worked, because of this issue #3 
First I thought that was because of lack ios and android directories. But when You said that it's working for You when You copy plugin files into main project package, I did some tests, and realized that when I remove those dependencies in yours plugin .yaml file, then project is compiling with success. And it makes sense, because You are not using any native SDKs, and with those dependencies included in yaml file builder was trying to use them, and then error was appear.

PR #5, opened 2018-10-12
Remove parse from endpoints path

PiotrWpl:
"parse" shouldn't be included in endpoint path, because if someone have domain dedicated just for Parse service (Parse is installed in root dir), this plugin will not work.

### DIFF
--- a/lib/parse_object.dart
+++ b/lib/parse_object.dart
@@ -13,7 +13,7 @@ class ParseObject implements ParseBaseObject {
   String get objectId => objectData['objectId'];
 
   ParseObject(this.className) {
-    path = "/parse/classes/$className";
+    path = "/classes/$className";
   }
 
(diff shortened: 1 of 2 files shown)


PiotrWpl:
For example in my company we have a dedicated domain for Parse service. If in this plugin "parse" will be hardcoded in path, people like I, will not be able to use it.
Whats more, when I'm using Parse SDK for Android or iOS, there they also not hardcode "parse" in path, You as a developer are responsible to add it to yours url if you have it configured like this, please check:
* https://github.com/parse-community/Parse-SDK-Android#setup
* for iOS I don't have an example, but I can ensure You that it's works the same as in android

phillwiggins:
Apologies, it's been a long day. I see my mistake. Will correct now.

PR #6, opened 2018-10-12
ParseData() should be initialized

PiotrWpl:
Because data variable was never initialized, it was always null, and getting values like "masterKey" or "applicationId" will never work correctly. So, we need to initialize it first.


phillwiggins:
Already done :-)

PR #7, opened 2018-10-12
Fix for not working user login & creation

PiotrWpl:
I had added:
* missing body for user creation request
* changed URI creation for login, because previous one didn't worked. Full url was setted as a path
* this commit also include changes from this pull request #5


phillwiggins:
Fixed

PR #60, opened 2019-01-28
set client key instead master key

pcegarra:
changed to use variable clientkey instead master key


pcegarra:
With this hotfix, the client key working perfectly

phillwiggins:
Perfect, apologies... Trying to do too many things at once!

PR #66, opened 2019-02-01
miss typo in get specific object by id

Saifallak:
it should be getObject not get


phillwiggins:
Question... Do they both work?

Issue #36, opened 2019-01-17
Does the SDK support ParseFile yet?


thphuc:
ParseFile is an important object of Parse platform. Do we support it yet?

phillwiggins:
Hey Tran

We have created ParseFile as a base object but it doesn't upload and download yet. Shouldn't take long to implement. That will be in a near release.

Thanks

thphuc:
That's great! @phillwiggins

phillwiggins:
I'll update this thread when I manage to get that working!

phillwiggins:
Added support. Really needs some testing though. Our current server doesn't have File Storage support. Please can you confirm?

ScanEat:
I tested but there is en error :

```
E/flutter (29953): [ERROR:flutter/shell/common/shell.cc(184)] Dart Error: Unhandled exception:
E/flutter (29953): NoSuchMethodError: The getter 'path' was called on null.
E/flutter (29953): Receiver: null
E/flutter (29953): Tried calling: path
E/flutter (29953): #0      Object.noSuchMethod (dart:core/runtime/libobject_patch.dart:50:5)
E/flutter (29953): #1      new ParseFile (file:///Users/toto/Desktop/flutter_parse_sdk-release-1.0.9/lib/src/objects/parse_file.dart:31:42)
E/flutter (29953): #2      parseDecode (file:///Users/toto/Desktop/flutter_parse_sdk-release-1.0.9/lib/src/utils/parse_decoder.dart:67:18)
E/flutter (29953): #3      ParseBase.fromJson.<anonymous closure> (file:///Users/toto/Desktop/flutter_parse_sdk-release-1.0.9/lib/src/objects/parse_base.dart:74:32)
E/flutter (29953): #4      __InternalLinkedHashMap&_HashVMBase&MapMixin&_LinkedHashMapMixin.forEach (dart:collection/runtime/libcompact_hash.dart:370:8)
E/flutter (29953): #5      ParseBase.fromJson (file:///Users/toto/Desktop/flutter_parse_sdk-release-1.0.9/lib/src/objects/parse_base.dart:64:16)
E/flutter (29953): #6      DietPlan.clone (package:flutter_plugin_example/diet_plan.dart:13:49)
E/flutter (29953): #7      ParseResponse._handleSingleResult (file:///Users/toto/Desktop/flutter_parse_sdk-release-1.0.9/lib/src/objects/parse_response.dart:98:49)
[...]

PR #67, opened 2019-02-02
Fix User objects not saving, Add support for Parse Files, Anonymous Login

klanmiko:
Hi Phil,

I am using your sdk to develop an application for our hackathon participants. I ran over a few issues that I fixed and I thought it woud be a good idea to contribute back to this project. Here are my changes:

* Fixed changes on User object not syncing to server due to missing session token when accessing API
* Added the ability to login anonymously by providing "AUTH_DATA" POST data with a unique UUID. Can be extended to allow for facebook and twitter auth in the future
* Added the ability to create a Parse file with a url and name and download the file to temporary storage
* Fixed bug with ParseFile initialized to null _file value.

Let me know if these need work

### DIFF
--- a/lib/parse.dart
+++ b/lib/parse.dart
@@ -10,6 +10,8 @@ import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:shared_preferences/shared_preferences.dart';
 import 'package:web_socket_channel/io.dart';
+import 'package:uuid/uuid.dart';
+import 'package:path_provider/path_provider.dart';
 
 part 'src/base/parse_constants.dart';
 
--- a/lib/src/base/parse_constants.dart
+++ b/lib/src/base/parse_constants.dart
@@ -7,6 +7,7 @@ const String keyLibraryName= 'Flutter Parse SDK';
 // End Points
 const String keyEndPointUserName = '/users/me';
 const String keyEndPointLogin = '/login';
+const String keyEndPointUsers = '/users';
 const String keyEndPointVerificationEmail = '/verificationEmailRequest';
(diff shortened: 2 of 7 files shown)


klanmiko:
helps with #36

ScanEat:
Hi klanmiko, are you sure parse files works ?

Because I tested with a new "picture" column with png file.

And when I print the payload, I have this sample:

```
"picture":"{\"__type\": \"Pointer\", \"className\": \"File\", \"objectId\": \"null\"}"}
```

phillwiggins:
Amazing work. 

I'm happy to merge and will test tonight to ensure the file upload is working correctly. 

Thanks!

klanmiko:
Thanks @phillwiggins! Thanks for making this library.

@ScanEat yes it works. The reason you see that is because the parse_encoder.dart encodes the object as a pointer and not a file. I think it is done this way for a specific reason...? to send to the server? Otherwise the print statements for the ParseUser and ParseFile, ParseGeo should be before that

ScanEat:
Ok @klanmiko but can you show us a code sample to retrieve the picture.url from database ?

rodrigosmarques:
Hi Guys,
Have you checked the ParseFile upload is working?
Do you have any sample code to send me?

phranck:
@klanmiko You added support for anon login, which is great! 😊 However, I wonder how to check if the `currentUser` is an anonymous user? Do you have a suggestion?

Thanks in advance!

danibjor:
@phranck you can check if 
```
var user = await ParseUser.currentUser();
if (user != null && user['userData'] != null && user['userData']['anonymous'] != null) { 
  // do anonymous user stuff
}
```

PR #68, opened 2019-02-02
Fix Parse User Login

FahimF:
Fixed Parse login requests not working due to incorrect request type


phillwiggins:
I'll double check this tonight. The login function has been working correctly as far as I am aware.

PR #86, opened 2019-02-24
Handle sessionToken correctly in ParseUser

chrbayer:
It is now possible to use the stored user with the stored sessionToken to autologin agian.


phillwiggins:
Great work.

Thank you Chris!

phillwiggins:
Great work.

Thank you Chris!

PR #91, opened 2019-03-03
v1.0.13 - Added full bool

phillwiggins:


phillwiggins:
Added full bool to correct json conversion

PR #95, opened 2019-03-05
ParseInstallation implementation

IonAgorria:
This adds ParseInstallation class, one issue I have found is getting the TimeZone from flutter in same way that Parse Server excepts (Europe/Madrid). 

Also an dependency to deviceinfo was added to get Locale value from device using platform code, a better approach would be to add platform specific code in this library instead of using external libraries. This could also solve the problem with TimeZone as we could simply copy the way iOS/Android handles it.

### DIFF
--- a/lib/parse_server_sdk.dart
+++ b/lib/parse_server_sdk.dart
@@ -13,6 +13,8 @@ import 'package:shared_preferences/shared_preferences.dart';
 import 'package:web_socket_channel/io.dart';
 import 'package:uuid/uuid.dart';
 import 'package:path_provider/path_provider.dart';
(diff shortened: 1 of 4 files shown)


rodrigosmarques:
How to use ACL. I noticed that get and set methods were implemented. Can you give an example?
The ACL could be implemented directly in ParseObject.

Thank you.

phillwiggins:
Hey

ACL isn't correctly implemented at the moment. You would need to manually
create an ACL object via JSON and send it along with your object. That's
close on my roadmap for this library.

rodrigosmarques:
OK. Thanks phillwiggins.

PR #130, opened 2019-03-18
Add fetch for ParseObject

IonAgorria:
This uses the standard fetch method from other sdks and eases the usage of pointers such as objA.get("pointerB").fetch();


phillwiggins:
Thank you very much, I will close this PR but I have added the code to the v1.0.16 branch!

Thank you!

PR #136, opened 2019-03-21
Setting installationId in _Session object during sign-up

mregandla:
I was needing this change during my app development. Feel free to close this pull request, if you find any issues.

You can refer the following link for more info.
https://github.com/parse-community/parse-server/issues/1097


mregandla:
Thanks for the review Phill. I added a null check and modified the code to prevent unnecessary entry in _Session object when installationId is null.

I am new to Dart. If you think any other changes required, feel free to suggest.

PR #143, opened 2019-04-01
Fix lost lat, lng data

danibjor:
Update so SubClass of ParseObject uses internal Map for value storage. Changes in v1.0.17 breaks ParseGeoPoint. This is tested and seems to fix the issue.


phillwiggins:
Brilliant, thanks.

PR #150, opened 2019-04-13
BugFix LiveQuery - update readme.md

rodrigosmarques:
- BugFix LiveQuery 
- Update README.md -> Reorganization of sessions and inclusion of documentation about LiveQuery

### DIFF
--- a/lib/src/network/parse_live_query.dart
+++ b/lib/src/network/parse_live_query.dart
@@ -1,52 +1,160 @@
 part of flutter_parse_sdk;
 
-/// Still under development
+enum LiveQueryEvent { create, enter, update, leave, delete, error }
+
 class LiveQuery {
+  LiveQuery({bool debug, ParseHTTPClient client, bool autoSendSessionId}) {
+    _client = client ??
+        ParseHTTPClient(
+            sendSessionId:
+                autoSendSessionId ?? ParseCoreData().autoSendSessionId,
+            securityContext: ParseCoreData().securityContext);
 
-  LiveQuery(ParseHTTPClient client) : client = client {
-    connectMessage = <String, String>{
+    _debug = isDebugEnabled(objectLevelDebug: debug);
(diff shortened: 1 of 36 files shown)


rodrigosmarques:
I appreciate if you can do a review in the text. My native language is not English and may have some error in my changes in README.md

phillwiggins:
Wow, great work!

rodrigosmarques:
Thank You.

I have not had time to update the sample application. I'll try to do.

But following the guidance of the documentation has no error.

lib/src/network/parse_live_query.dart
```diff
@@ -1,52 +1,180 @@
...
+        }
+        return Future.value(handleException(error, ParseApiRQ.liveQuery, _debug, _className));
+      });
+
+      //The connect message is sent from a client to the LiveQuery server.
+      //It should be the first message sent from a client after the WebSocket connection is established.
+      _connectMessage = {
+        'op': 'connect',
+        'applicationId': _client.data.applicationId,
+        'clientKey': _client.data.clientKey
```
danibjor: As `clientKey` is optional, this line throws an error if `clientId` is not set in `Parse().initialize( ... )`
```suggestion
        'clientKey': _client.data.clientKey ?? ''
```

rodrigosmarques: it is true.  I had not thought of that.  I have to review the whole logic because the key can not be sent empty.

danibjor: I tested setting an empty string as key when initializing Parse, that works just fine. So the fix suggested should do the trick.

rodrigosmarques: OK. I'm making the change now.

rodrigosmarques: Done @phillwiggins

phillwiggins:
Brilliant!

PR #166, opened 2019-04-30
remove shared preference dependance 

WahdanZ:
create CoreStore interface to allow different implements for local storage 
implement default local store using sembast and with encryption using XXTEA algorithm
add desktop support an example project


WahdanZ:
for ParseInstallation I think I will be useful to remove it and added it in a sperate package for flutter to also allow support notification

phillwiggins:
Amazing work! 

I will test this out tonight. Looking good code wise.

PR #199, opened 2019-06-17
Fix save

yulingtianxia:
`ParseGeoPoint` should be simple. 
This PR could fix some issue about saving objects. (https://github.com/phillwiggins/flutter_parse_sdk/pull/195#issuecomment-502511399)


rodrigosmarques:
@yulingtianxia 
ParseGeoPoint is not a Parse object is a data type and therefore does not extend ParseObject Class.

It is not necessary to change the parse_encode functions to put .toJson () because this method is already called by default.

This code is working properly today. I have made all the adjustments and tests.

I'm building an app that uses all these functions and everything is ok.

yulingtianxia:
@rodrigosmarques 
OMG, it's a mistake. I'll delete the two lines.

rodrigosmarques:
No need toJson. When you call the jsonEncode method it is already called by default in value

https://api.dartlang.org/stable/2.4.0/dart-convert/jsonEncode.html

yulingtianxia:
@rodrigosmarques 
Thank you for your correction!
There is something wrong with my git merge. It's not intensional to add toJson. I had deleted it.

PR #216, opened 2019-07-10
fix some issues

zenz:
fix liveQuery.subscribe issue.
fix installation issue.

It's now working with my livequery sample.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -64,7 +64,7 @@ class _MyAppState extends State<MyApp> {
   Future<void> initData() async {
     // Initialize repository
     await initRepository();
-    final CoreStore coreStore = await initCoreStore();
+    // final CoreStore coreStore = await initCoreStore();
 
     // Initialize parse
     await Parse().initialize(keyParseApplicationId, keyParseServerUrl,
@@ -372,7 +372,6 @@ class _MyAppState extends State<MyApp> {
     dietPlanRepo ??= DietPlanRepository.init(await getDB());
     userRepo ??= UserRepository.init(await getDB());
   }
-
 
   /// Available options:
   /// SharedPreferences - Not secure but will work with older versions of SDK - CoreStoreSharedPrefsImpl
--- a/lib/src/objects/parse_base.dart
+++ b/lib/src/objects/parse_base.dart
@@ -3,7 +3,7 @@ part of flutter_parse_sdk;
 abstract class ParseBase {
   String parseClassName;
   Type type;
-  bool _dirty = false; // reserved property
+  final bool _dirty = false; // reserved property
   final Map<String, dynamic> _unsavedChanges = Map<String, dynamic>();
   final Map<String, dynamic> _savingChanges = Map<String, dynamic>();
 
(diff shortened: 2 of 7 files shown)


example/lib/main.dart
```diff
@@ -64,7 +64,7 @@ class _MyAppState extends State<MyApp> {
   Future<void> initData() async {
     // Initialize repository
     await initRepository();
-    final CoreStore coreStore = await initCoreStore();
+    // final CoreStore coreStore = await initCoreStore();
```
phillwiggins: Why has this been removed?

lib/src/objects/parse_base.dart
```diff
@@ -3,7 +3,7 @@ part of flutter_parse_sdk;
 abstract class ParseBase {
   String parseClassName;
   Type type;
-  bool _dirty = false; // reserved property
+  final bool _dirty = false; // reserved property
```
phillwiggins: This feels weird. Why is this final. 

This might not be related to what you are doing here, but does seem weird that we even set, _dirty and finalise it.

lib/src/objects/parse_base.dart
```diff
@@ -100,6 +102,7 @@ abstract class ParseBase {
       map.remove(keyVarUpdatedAt);
       map.remove(keyVarClassName);
       //map.remove(keyVarAcl);
+      map.remove(keyVarObjectId);
```
phillwiggins: Are we sure this is correct for all calls? Seems strange that it's there in the first place. Have you ran tests that everything else works with this removed? Rather than just installation?

example/lib/main.dart
zenz: bwcause it is not used yet. so I comment it out.

lib/src/objects/parse_base.dart
zenz: currently in your code you didnt modify it. so I just follow dart's lint to modify it to final.

lib/src/objects/parse_base.dart
zenz: I do all the test in example. yes.

phillwiggins: So this works when saving?

I'm not sure saving an object would work anymore. It will create a new object.

example/lib/main.dart
phillwiggins: This seems like a big and will need to be investigated further.

phillwiggins:
Apologies, I've just noticed as well, we are ok branch release/1.0.23 now.

Please can you make a PR there instead.

zenz:
ok, I will check it tomorrow. it's time to sleep here.

phillwiggins:
Ah apologies! Thanks for your effort by the way.

lib/src/objects/parse_base.dart
rodrigosmarques: @phillwiggins 
I'm seeing that they removed the ACL from ParseObject, the code was commented out. When was it done? I've added this functionality to lib and am using this in version 21 on a large project and have not had problem.

rodrigosmarques:
@phillwiggins 
I can not use version 22. It did not pass my test app. Once the merge is done in version 23, I'm going to retake the tests and would like to know why the ACL was removed.

zenz:
@rodrigosmarques 
Please try this:
```
parse_server_sdk:
   git:
     url: git@github.com:zenz/flutter_parse_sdk.git
     ref: 63da5b4243d213824541a35d71c72e82f57c6ac6
```

PR #218, opened 2019-07-10
same fix for version 1.0.23

zenz:
fix installation issue
fix liveQuery.subscribe issue


zenz:
And here's the samples to test this SDK v 1.0.23 with livequery.
It just modified from sdk example code to make sure how it works.

PR #223, opened 2019-07-14
Fix crash on Flutter Web

rostopira:


WahdanZ:
Could you show us how to use the sdk with flutter web ?

rostopira:
@WahdanZ it's basically straightforward with some caveats
```
flutter channel master
flutter upgrade --force
flutter run -d chrome
```
LiveQuery not working currently (I'm working on it)
And for `CoreStore` I'm using stub (WIP as well)
SecurityContext is unsupported, throws on constructor, just don't use it
Also be aware, that it's doesn't support `dart:io` and will throw even on `Platform.isAndroid`
You can still use it like this:
```
if (identical(0, 0.0) {
  // I'm Flutter Web
  // Use 'dart:html' package
} else if (Platform.isAndroid) {
  // Use 'dart:io'
}
```

rostopira:
As I already said, it's WIP, I will provide more PR's for flutter web soon, just part by part, not all at once

PR #240, opened 2019-08-07
MergeWithPrevious change

WangPerryWPY:
The method of setAdd, setRemove, setAddAll, setRemoveAll, Increment, AddUnique, AddRelation, RemoveRelation can be merged between the same operation and some other operation. Implementation method reference iOS SDK. I do this by Implementing a tool class in "parse_merge.dart" and calling the "mergeWithPrevious" Method in the set method which implemented in parse_base.dart.

### DIFF
--- a/lib/src/objects/parse_base.dart
+++ b/lib/src/objects/parse_base.dart
@@ -191,12 +191,6 @@ abstract class ParseBase {
         if (_getObjectData()[key] == value) {
           return;
         }
-        /*if (!forceUpdate) {
-          if (value is Map<String, dynamic>) {
-            value = mergeWithPrevious(_getObjectData()[key], value);
-          }
-        }*/
-        // value = mergeWithPrevious(_unsavedChanges[key], value);
         _getObjectData()[key] =
             ParseMergeTool().mergeWithPrevious(_unsavedChanges[key], value);
       } else {
--- a/lib/src/objects/parse_base.dart
+++ b/lib/src/objects/parse_base.dart
@@ -191,13 +191,18 @@ abstract class ParseBase {
         if (_getObjectData()[key] == value) {
           return;
         }
-        if (forceUpdate) {
-          _getObjectData()[key] = value;
-        }
+        /*if (!forceUpdate) {
+          if (value is Map<String, dynamic>) {
+            value = mergeWithPrevious(_getObjectData()[key], value);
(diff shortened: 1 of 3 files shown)


lib/src/objects/parse_base.dart
```diff
@@ -191,13 +191,18 @@ abstract class ParseBase {
         if (_getObjectData()[key] == value) {
           return;
         }
-        if (forceUpdate) {
-          _getObjectData()[key] = value;
-        }
+        /*if (!forceUpdate) {
```
phillwiggins: Please can you remove this commented out code.

phillwiggins:
Great work! 

Will merge once checks have finished

WangPerryWPY:
Yes, I have tested what I changed, and have removed the commented.
Thanks
 
 
------------------ Original ------------------

rodrigosmarques:
@WangPerryWPY 
I did not understand the change.
These methods already work this way today.
Can you explain the advantage of this change over the previous implementation?

WangPerryWPY:
Before my change, if I setAdd('Test', 1); then setAdd('Test', 2); the first operation will be covered by the last. But after my Change, these operations will become setAdd('Test', [1, 2]); just like all the other parse SDK documents. And setRemove, Increment, setAddUnique is the same like this. 
 
------------------ Original ------------------
From: "Rodrigo de Souza Marques"; 
Date: 2019年8月7日(星期三) 晚上6:24
To: "phillwiggins/flutter_parse_sdk"; 
Cc: "WangPerryWPY"; "Mention"; 
Subject: Re: [phillwiggins/flutter_parse_sdk] MergeWithPrevious change (#240)

 

@WangPerryWPY
 I did not understand the change.
 These methods already work this way today.
 Can you explain the advantage of this change over the previous implementation?

pastordee:
Hi the same thing needs to be done with queryBuilder as well, e.g if you say 
.. orderByDescending('online')
.. orderByDescending('last-seen');

to order online first and then last seen, 
only the online(first one) works, the last seen is not in order but on iOS SDK and ANDROID SDK it works

rodrigosmarques:
@pastordee 
I had not tested it.
I'm doing a PR with the fix

rodrigosmarques:
@phillwiggins and @pastordee 
#241

PR #246, opened 2019-08-09
Add LiveQuery support for Flutter Web

rostopira:
Adds LiveQuery support for Flutter Web using WebSocket implementation from `dart:html` package


rostopira:
CI will fail most likely with `Undefined class WebSocket` error
I dunno, how to silence it

PR #262, opened 2019-08-29
Bug fix.

yulingtianxia:
Bug fix: 'set' method should force update by default. If someone change a property of object from 'get' method and then 'set' it in ParseObject, it should be dirty.


phillwiggins:
Good work! Thanks

PR #263, opened 2019-09-02
Fix About Retrieving Sessions

hellokidder:
hi!
I find a BUG when I want retrieving sessions,and I Fix it
before：
```
url = http://171.11.60.130/parse/sessions/me
```
now:
```
url = http://171.11.60.130:8000/parse/sessions/me
```

look forward to your reply，have a good day!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .DS_Store
 .dart_tool/
-
+.history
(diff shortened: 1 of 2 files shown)


phillwiggins:
Great work, thank you!

PR #272, opened 2019-09-12
Added headers to ParseUser.destroy which are also passed to ParseHTTPClient.delete

matskosan:
I needed to send headers to cloud triggers beforeDelete and afterDelete.

### DIFF
--- a/lib/src/objects/parse_user.dart
+++ b/lib/src/objects/parse_user.dart
@@ -315,11 +315,11 @@ class ParseUser extends ParseObject implements ParseCloneable {
   }
 
   /// Removes a user from Parse Server locally and online
-  Future<ParseResponse> destroy() async {
+  Future<ParseResponse> destroy({Map<String, String> headers}) async {
     if (objectId != null) {
       try {
         final Uri url = getSanitisedUri(_client, '$_path/$objectId');
-        final Response response = await _client.delete(url);
+        final Response response = await _client.delete(url, headers: headers);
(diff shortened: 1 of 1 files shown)


matskosan:
I just send a custom text header 'x-password-confirmation' with the user password that I check in 'beforeDelete' trigger on Parse.User to make sure this is the actual owner who wants do delete the account and not someone else. This surely can be used to send any other headers.

davimacedo:
I think that's not the best way to be done. In order to create this feature, I'd recommend creating a cloud code function that receives the password, do the verification, and, in case of success, delete the user. Then you can disable the delete access from the _User class CLP so you can make sure that users will only be deleted through this cloud code function.

I am not sure if it is a good idea to allow the developers to pass arbitrary headers through the SDK.

matskosan:
The 'delete' cloud function was my initial approach. However, I also felt that having this check in a trigger feels more 'natural', as the functionality is already there. Maybe it makes sense to change parse server in a way that deleting some objects would require owner's password?

I am fine cancelling the pull request though and creating the cloud function. Is it how everyone else does this?

matskosan:
Do you know if there is a way to set up CLP using code so that is stays consistent between deployments?

davimacedo:
@matskosan I think that the best way is really the usage of cloud code. In terms of CLP, you can take a look in this issue: https://github.com/parse-community/parse-server/issues/5888

PR #290, opened 2019-12-07
fix dependency conflict on Sembast in example

jwfing:


phillwiggins:
Hey, I apologies, I didn't see this! Please can you point this to branch 'release/1.0.25'. I will merge it as soon as. Thanks

zenz:
I would suggest a fresh update but not only fix dependency. 
since updated to Sembast v2, many old definition was deprecated.

PR #294, opened 2019-12-17
example app migration to use Sembast v2

mregandla:


phillwiggins:
Great work, thank you!

PR #295, opened 2019-12-17
cleanup and fixed a broken code

mregandla:
1. Updated .gitignore
2. Removed unnecessary files from the repo
3. Fixed broken code


.gitignore
```diff
@@ -6,9 +6,30 @@
...
 .example/lib/application_constants.dart
 
+# IntelliJ related
+*.iml
+*.ipr
+*.iws
+.idea/
+
+# Android related
+gradle.properties
```
mregandla: I think, gradle.properties should not be ignored. @phillwiggins Do you have any comments? Let me know if I need to fix this.

mregandla: Never mind, this is fixed now.

phillwiggins:
Great work! All looks good.

.gitignore
phillwiggins: Sorry for the delay, I've been away. Yes, you've done the correct thing by not adding gradle.properties to gitignote. Thanks

PR #306, opened 2020-02-15
Fixed the Parse().initialize's return value

th3brink:
Parse().initialize returns a new instantiation of the Parse class, but it should return the current instance.


phillwiggins:
Good spotting!

Please can you close this MR and create a new MR labelled branch release/1.0.26.

Thanks in advance

PR #308, opened 2020-02-18
Added an example of how to update existing object values

rommyarb:
I have been struggling with finding a way to update the values of the existing object. After many trials and errors, I finally found it.

I added the example. I think it will be useful for anyone who is new to this SDK. Maybe my example needs some correction.

Thank you.


phillwiggins:
Hey @rommyarb 

Thanks for the addition. Please can you close this MR and add your work to  the branch labelled "release/1.0.26". 

Thanks in advance!

PR #310, opened 2020-02-18
Specifying return type of get<T> as T instead of dynamic

BaranMichal25:
Small change, but I think it's a good practice to return a concrete type instead of dynamic when we can.

Additionally, for developers who have `implicit-casts` set to `false` in their analysys_options.yaml, from this point on they won't have to write code like this:
`String name = get<String>('name') as String;`
but instead will be able to do just:
`String name = get<String>('name');`

Thanks for making this package btw! :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
+## 1.0.26
+
 ## 1.0.25
+Update dependencies
 
 ## 1.0.24
 Fixed lint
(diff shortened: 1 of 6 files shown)


phillwiggins:
Hey @BaranMichal25 

Great suggestion. Have you tried and testing this in anger?

Quick one, are you able to close this MR and create a new MR pointing to branch labelled "release/1.0.26"? I will merge it there, give it some testing too! 

Cheers

BaranMichal25:
Hey @phillwiggins 

I changed here the target branch to `release/1.0.26`.
I also tested it in our app (multiple tables, requests, mappers) and everything works fine 👍

phillwiggins:
Excellent, good effort!

Issue #261, opened 2019-08-28
Provide an event or a way to detect when connection is lost while using LiveQuery


andyalbert:

mdc555:
Same here.
This would be great! I was looking for this...

qwertyway:
One more wish for this

allanwolski:
Same here.

krayony:
Same.

adrianaxente:
Same

andyalbert:
The issue is solved & merged by PR #314

PR #313, opened 2020-03-02
Live query connection stream

maaeps:
This change provides "an event or a way to detect when connection is lost while using LiveQuery" as requested in #261 
Please consider merging this change.


phillwiggins:
Hey, great work. 

Please can you close this MR and point it to 'release/1.0.26'? 

Thanks in advance!

Issue #315, opened 2020-03-02
LiveQuery: "already connecting"


maaeps:
Currently "_connecting" (parse_live_query.dart) is not properly handled properly.
This results in a state where the client can't establish a new connection to the server.

**How to reproduce:**
1. connect to the livequery-server
2. disconnect the client from the internet
3. reconnect the client to the internet
4. you won't be able to reconnect to to the livequery-server
(Tested on an android emulator)

**The Problem:**
- the IOWebSocketChannel crashes and calls onDone
- _userDisconnected is false => reconnect is called
- _connecting is set to true
- WebSocket.connect fails => catch called

reactivate wifi => onConnectivityChanged fires event
- reconnect is called
- **BUT** _connecting is still true => _connect is canceled

Similar problems occur when the server crashes.
(and at some other times? (not sure))

**How to fix it:**
My suggestion is:
- set _connecting to false in the catch block
- adding a timeout to WebSocket.connect (in case the server is down)
- set _connecting to false after WebSocket.connect is done
- remove "_connecting = false;" inside _handleMessage
- adding some kind of delayed reconnect inside the catch block (in _connect)

**Any other ideas?**

maaeps:
#316 is merged

PR #316, opened 2020-03-03
Livequery reconnecting controller

maaeps:
Should fix the issue described in #315 and improve the reconnecting after a connection loss to the live-query-server.

Please consider merging.

Cheers


phillwiggins:
Looks good, great work!

Issue #213, opened 2019-07-07
Does the SDK support OR query?


thphuc:

rodrigosmarques:
does not support.  I even started to implement, but I did not finish

thphuc:
Hi @rodrigosmarques 
Hope you are doing well.
How is it going?
I may have some time to help on this, please let me know. Thanks!

WahdanZ:
@rodrigosmarques  it there any update regarding this one I can do some help also

rodrigosmarques:
Hi @WahdanZ .
I didn't have time to go back in this feature.
If you have conditions, you can build

maaeps:
Hi,
I am currently working on such an OR Query.
In the Android SDK, they throw an error, if one of the subquerys uses "selectKeys"/"dontSelectKeys".

Does anyone know, if the Parse-Server does not support this or if it's just missing feature in the android implementation?

maaeps:
This feature got added in #317 without caring about "selectKeys"/"dontSelectKeys".

Loudsa:
`
     QueryBuilder query = QueryBuilder.or(ParseObject("Test"), [
        QueryBuilder(ParseObject("Test"))..whereStartsWith("order", 1),
        QueryBuilder(ParseObject("Test"))..whereStartsWith("number", 'test-1'),
      ]);
`
I think you should say this

PR #317, opened 2020-03-04
Parsequery OR

maaeps:
Added a second constructor to ParseQuery for creating OR Queries.
#213 

Tested with basic queries.
Works with live query, too.

### DIFF
--- a/lib/src/network/parse_query.dart
+++ b/lib/src/network/parse_query.dart
@@ -6,14 +6,15 @@ class QueryBuilder<T extends ParseObject> {
   QueryBuilder(this.object) : super();
 
   QueryBuilder.or(this.object, List<QueryBuilder<T>> list) {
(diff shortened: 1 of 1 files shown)


phillwiggins:
That looks really promising! Good effort. I tried looking at this a while ago but ended up giving up. That looks like a promising solution.

RodrigoSMarques:
@maaeps 
Very good. I also started to develop this in the past, but due to lack of time I was unable to progress.

RodrigoSMarques:
It is on my list to complete the rest of the functions with GeoPoint. Some are still missing. I got involved with other plugins, but I intend to go back and finish.

maaeps:
Thanks for merging.
The parsing of the subqueries is a little goofy.
Maybe someone will find a more pretty solution.

Issue #319, opened 2020-03-05
Proper QueryBuilder copy


maaeps:
Is this already implemented?
Can't find it.

PR #320, opened 2020-03-05
added QueryBuilder.copy(QueryBuilder<T> query)

maaeps:
I could not find any way to copy a QueryBuilder.

So I added a constructor that takes an QueryBuilder as a base.
This is a deep copy.
Tested only with basic queries.

#319

### DIFF
--- a/lib/src/network/parse_query.dart
+++ b/lib/src/network/parse_query.dart
@@ -17,6 +17,16 @@ class QueryBuilder<T extends ParseObject> {
     }
   }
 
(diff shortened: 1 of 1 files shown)


maaeps:
Just added the second and third commit to rerun the check.

maaeps:
I have no idea whats causing the tests to fail.

Can anyone give me a hint?

maaeps:
Seems like [sembast](https://pub.dev/packages/sembast) got an update.

PR #334, opened 2020-03-12
LiveQuery: list is null fixes

maaeps:
When no object fits the query, the the returned list is null.
This should fix some NullPointer errors.


maaeps:
I solved the conflict.

Issue #364, opened 2020-04-28
Parse user updated on server but not locally


GursheeshSingh:
I think this a strange behaviour in ParseUser

```
ParseUser currentUser = await ParseUser.currentUser();
currentUser.set("turnOnNotifications", newValue);
var response = await currentUser.update();
if (response.success) {
    ParseUser currentUser = await ParseUser.currentUser();
    currentUser.get("turnOnNotifications");
    return;
}
```

Here when I change the value of current user and update it. It successfully updates the value to the sever and the response also gives the updated value

But when I access it again, it is still unchanged, I think the update only happens on the server and not locally.

This is quite frustrating as it causes a lot of bugs and does not help in the program flow.

maaeps:
Seems like this is related to #256

GursheeshSingh:
Yes I think so.

The user is logged in already and it successfully gets updated on the server. But not locally, it seems like a SDK issue.

maaeps:
I've done some testing, too.
Before copying your sample code, I used `.save()` to save the, everything worked as expected.
Then I changed `.save()` to `.update()`, this resulted in the same problems as described you.
By logging out and back in the user object  was updated.

I think by calling update, the SDK saves changes successfully but does not write the UserObject to the CoreStore properly.

I will try to fix this tomorrow.

GursheeshSingh:
Does the .save() work for you?

.save() and .update() don't update the local storage for me.

maaeps:
For me `.save()`works just fine.
Which corestore are you using?

GursheeshSingh:
Seems like its working with save() and not update().
I am using Shared Preferences

Can we use Flutter Secure storage?

GursheeshSingh:
Does save() work the same if I make some changes and save it?

maaeps:
I am actually not quite sure what `.update()` does.
`.save()` should just save your changes to the server.
Have you tried the changes merged with #368?

GursheeshSingh:
I will try them out and let you know.

GursheeshSingh:
How do I specifically use this version pull request in pubspec.yaml?

phillwiggins:
Something like this do you mean?

`parse_server_sdk:
    git:
      url: git://github.com/phillwiggins/flutter_parse_sdk.git
      ref: release/1.0.26`

GursheeshSingh:
Thank you.
```

  parse_server_sdk:
    git:
[...]

Issue #350, opened 2020-04-07
Should we add registerSubclass?


maaeps:
This issue is part of #194 

**Should we add registerSubclass to match the Android/IOS SDK?**

**Pros:**
- usage of custom ParseObject classes would be much simpler

**Cons:**
- huge change
- existing code might break by updating the SDK

phillwiggins:
Can you explain the pros in a bit more depth?

maaeps:
Often the user can provide a copy-object. This is used to create new objects based on a custom ParseObject class.
By holding this copy-objects in the background, we would be able to create every object based on the right sub class.
As far as I know, included objects in a query are always of type ParseObject. This would be avoidable, too.

In summary: the user would not have to care about providing such copy-objects.

phillwiggins:
Apologies for the late reply. I still don't actually understand this. Do you have you any examples where this is done already? I understand the Parse Android SDK does but as far as I'm aware there's not too much benefit when using Flutter as we don't have reflection to iterate through class structures?

maaeps:
I did was not working on this for a while, too.
But I've started implementing this feature now.
To get an idea what I am talking about, you can checkout my [branch](https://github.com/maaeps/Parse-SDK-Flutter/tree/registerSubclass).
I you run something like this after the `Parse().initialize();`
```
ParseCoreData.instance
      ..registerSubClass("Test", () => Test1())
      ..registerSubClass("Test2", () => Test2())
      ..registerSubClass("Test3", () => Test3());
```

The SDK will use your custom classes for objects with the classnames 'Test', 'Test2' and 'Test3'.

This should allow you to do something like this:
`Test2 test = object.get<Test2>('test');`
Without this changes a error gets thrown: `type 'ParseObject' is not a subtype of type 'Test2'`

maaeps:
...adding a Constructor to QueryBuilder where the object the registered sub-classes get uses automatically would be useful, too.

phillwiggins:
I really like this idea. Seems very logical. Did you have any luck
implementing it?

maaeps:
Well, I think something as simple as [this](https://github.com/maaeps/Parse-SDK-Flutter/commit/f157becbdc042bcb76b51c9ea6462d61c994c56c) should work.

maaeps:
@phillwiggins 
I've added some more changes to [my branch](https://github.com/maaeps/Parse-SDK-Flutter/tree/registerSubclass).
[...]

Issue #256, opened 2019-08-23
Cannot update custom value in User class


Tridie2000:
My User class has a custom property (points). I'm trying to update this property. According to a print statement I added to the code it seems to work except I don't see any change on the server side. 

```
ParseUser user = await ParseUser.currentUser() as ParseUser;
if (user != null) {
   ParseUser incrementUser = ParseUser.forQuery()..getObject(user.objectId);
   incrementUser.set('points', (user.get('points') + 1));
   await incrementUser.save();
   print('save: $incrementUser');
}

**Comment by Tridie2000** [comment]
I also tried this piece of code and the response is successful but the data doesn't update on the server. What is going on here?

```
ParseUser user = await ParseUser.currentUser() as ParseUser;
    if (user != null) {
      user.set('points', user.get('points') + 1);
      ParseResponse response = await user.save();
      if (response.success) {
        print('ok');
      } else {
        print(response.error);
      }
      print('save: $user');
    }

danibjor:
Check user.objectId. Is that set? Or is it null?

Tridie2000:
The user objectId is not null.

mdc555:
Any update on this??
Same here. I've tried several ways but it's not updating. But curiously, it was working fine and I didn't any change to my code....

mdc555:
The User doesn't get updated when it's logged in....

jbcaveman:
Did you check the logs for any permission errors? Most likely you need to specify using the master key for updating User object because Parse automatically generates tight ACLs for that class. Try:
`incrementUser.save(null, { useMasterKey: true });`

mdc555:
I've finally managed to get this working. I had to move that logic to a Cloud Code passing `masterKey: true` and everything is fine but I couldn't get it working straight from my app. I didn't get any error on my logs. Supposedly everything was OK but the user never updated.
I tried to logout and login back the user and at that point it worked but whenever I started my app that update never happened. A bit strange but now solved using Cloud Code.

jbcaveman:
Parse always generates tight ACLs on the User class, so you cannot modify a user in the client SDK unless that user is the authenticated currentUser. That said, you should always increment values on the server side rather than client side to avoid stale data issues and other conflicts. To save objects with tight ACLs server-side, you need to be sure to specify using the master key, like: 
`incrementUser.save(null, { useMasterKey: true });`

jbcaveman:
[...]

PR #368, opened 2020-04-29
RegisterSubClass and smaler changes

maaeps:
As described in https://github.com/parse-community/Parse-SDK-Flutter/issues/350#issuecomment-614071136, I have now worked with this RegisterSubClass feature for a while.
While testing, I haven't encountered a single bug.
I think, this feature is ready for merging.

For more details about RegisterSubClass, have a lock into the new README.

This PR also contains a Fix for a Bug, where CoreStore would not work when using the sembast implementation in an macos app.

And a fix for #364 / #256.

I have created one PR for this three changes. If this is a problem, I can create separated ones as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,24 @@ class DietPlan extends ParseObject implements ParseCloneable {
   
 ```
 
+When receiving an `ParseObject` from the SDK, you can often provide an instance of your custom object as an copy object.
+To always use your custom object class, you can register your subclass at the initialization of the SDK.
+```dart
+Parse().initialize(
+   ...,
+   registeredSubClassMap: <String, ParseObjectConstructor>{
+     'Diet_Plans': () => DietPlan(),
+   },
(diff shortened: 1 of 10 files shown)


phillwiggins:
Wow. I'm excited for this feature! I think this could answer a few of my problems in my own projects.

Just to confirn, is this a breaking change?

maaeps:
@phillwiggins 
Everything should work as before. Nothing should be broken.

ChineOnw:
When will this appear in pub.dev repository, since the newest version there is 1.0.26 ?

maaeps:
@ChineOnw 
There is, as far as I know, no plan, when 1.0.27 will be released to `pub.dev`.
@phillwiggins Might be able to give you a rough idea, when this will happen.

You can use 1.0.27 using this in your `pubspec.yaml`.
```
  parse_server_sdk:
    git:
      url: git://github.com/parse-community/Parse-SDK-Flutter.git
      ref: release/1.0.27
```

PR #376, opened 2020-05-18
ParseLiveElement

maaeps:
ParseLiveList listens on included objects. LiveQuery is currently not able to do this due to the server implementation.

As this is an useful feature, I added LiveElement.
ParseLiveElement is build on to of ParseLiveListElement and can be used like a mix of LiveQuery and ParseLiveListWidget.
You can get the current ParseObject through a stream.


phillwiggins:
Very nice!

PR #391, opened 2020-06-11
[web] Check web before Platform.isX

nstrelow:
Fixes `Unsupported operation: Platform._operatingSystem` as described in #385 .

dart:io 's Platform.isX is unsupported in web.

The recommended way is using kIsWeb from flutter foundations: https://github.com/flutter/flutter/issues/36126.

This does not enable sembast usage in web. It still needs the [sembast_web package](https://pub.dev/packages/sembast_web).

PR #392, opened 2020-06-11
[web] Check web before Platform.isX

nstrelow:
Fixes `Unsupported operation: Platform._operatingSystem` as described in #385 .

dart:io 's Platform.isX is unsupported in web.

The recommended way is using kIsWeb from flutter foundations: https://github.com/flutter/flutter/issues/36126.

This does not enable sembast usage in web. It still needs the [sembast_web package](https://pub.dev/packages/sembast_web).

(Same as #391 but against the right branch)


phillwiggins:
Great work. Thank you

PR #393, opened 2020-06-11
Untested: Sembast web

nstrelow:
Work in Progress: I encourage others to contribute and make their own attempt.

I added the necessary packages sembast_web and included the correct databaseFactory.

I have not really tested this, as we currently do not use Sembast.

I would appreciate, this being tested by in an app taking full advantage of Sembast.


nstrelow:
Should not include previous commit.

PR #394, opened 2020-06-11
WIP: Add sembast_web and load correct databaseFactoryWeb

nstrelow:
Work in progress: I would appreciate if this is used as a starting point to get sembast working on web.
Feel free to use these changes in your own PR to fix it 🚀 

I added the necessary packages sembast_web and included the correct databaseFactory.

I have not really tested this, as we currently do not use Sembast.

Currently it kind of freezes on User.loginAnonymous() with error:
```
Exception: MissingPluginException(No implementation found for method currentLocale on channel uk.spiralarm.flutter/devicelocale)
```


nstrelow:
We cannot use codec, which is needed for encryption. Would it still make sense to use it for it's other advantages?
Then we could simply remove the code/password/encryption part on web and add a disclaimer that sembast_web does not support it.

Or we encrypt/decrypt values before/after accessing sembast via a different encryption method.

nstrelow:
Has been completed in #436

PR #395, opened 2020-06-11
Make web compatible calls for locale and skip PackageInfo

nstrelow:
Adds to #392 to further improve web compatibility.
General web support issue: #385

Replaces following incompatible packages only for web:

#### Devicelocale.currentLocale (https://github.com/magnatronus/flutter-devicelocale/issues/9)
Replaced by `ui.window.locale.languageCode` from dart:ui package.
This outputs the language as ("en_US", "en_GB"), same as on Android with Devicelocale (Devicelocale returns - dashes for iOS).

#### PackageInfo (https://github.com/flutter/flutter/issues/46609)
Here I opted to remove all the properties coming from PackageInfo.
Parsing the yaml file and injecting them inside the app somehow could work, but I would rather wait and hope for PackageInfo to support web sometime in the future.

Pending Readme changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,15 @@ It's possible to add other parameters to work with your instance of Parse Server
 	coreStore: await CoreStoreSharedPrefsImp.getInstance()); // Local data storage method. Will use SharedPreferences instead of Sembast as an internal DB
 ```
 
+
+#### Early Web support
+Currently this requires adding `X-Parse-Installation-Id` as an allowed header to parse-server.
+When running directly via docker, set the env var `PARSE_SERVER_ALLOW_HEADERS=X-Parse-Installation-Id`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -49,6 +49,15 @@ It's possible to add other parameters to work with your instance of Parse Server
 	coreStore: await CoreStoreSharedPrefsImp.getInstance()); // Local data storage method. Will use SharedPreferences instead of Sembast as an internal DB
 ```
 
+
+#### Early Web support
+Currently this requires adding `X-Parse-Installation-Id` as an allowed header to parse-server.
+When running directly via docker, set the env var `PARSE_SERVER_ALLOW_HEADERS=X-Parse-Installation-Id`.
+When running via express, set [ParseServerOptions](https://parseplatform.org/parse-server/api/master/ParseServerOptions.html) `allowHeader: ['X-Parse-Installation-Id']`.
```
theblissprogrammer, line 56: It should be allowHeaders with an 's'.

PR #407, opened 2020-08-05
ParseLiveList: option to turn off lazy loading

maaeps:
The lazy loading of ParseLiveList is a feature that allows to display long lists as quick as possible.
But for short lists it takes some extra time.
This is a first implementation that allows to turn of lazy loading.
Setting `lazyLoading=false` should disable lazy loading.
In my (quick) tests everything worked as expected. However, further testing is highly appreciated.


maaeps:
@phillwiggins Wow that was fast, thanks for merging!

PR #430, opened 2020-08-29
Replace devicelocale by Platform.localeName call

nstrelow:
Devicelocale is not maintained a lot and currently throws errors for me when building a Android APK release.

It was only used in one place for the installation, and can be replaced by Platform.localeName (not working for Web, so keeping the alternate web call).
Now the package has one less dependency on a not greatly maintained package.

A throughout example or possibilities to get Locale: https://stackoverflow.com/a/62825776/1387765

Also replaces test (dart only), by flutter_test to fix tests.

### DIFF
--- a/lib/parse_server_sdk.dart
+++ b/lib/parse_server_sdk.dart
@@ -7,7 +7,6 @@ import 'dart:math';
 import 'dart:typed_data';
 import 'dart:ui' as ui;
 
-import 'package:devicelocale/devicelocale.dart';
 import 'package:http/http.dart';
 import 'package:http/io_client.dart';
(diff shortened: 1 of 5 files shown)


nstrelow:
Pipeline fails because of:

```
Error: cannot run without a dependency on "package:flutter_test". Ensure the
following lines are present in your pubspec.yaml:
dev_dependencies:
  flutter_test:
    sdk: flutter
```

Apparently tests now error if you use test, but have flutter as a dependency.
We need to add flutter_test

phillwiggins:
Good work! I'll fix the test situation very shortly. The lack of tests is probably the most worrying part 😬

PR #436, opened 2020-09-03
Add WEB-support for CoreStoreSembastImp

maaeps:
This is pretty much the same PR as @nstrelow's  #394.

Please have a look at  #394 for a detailed explanation of the problem.

I think, although it's not optimal,  not encrypting the database is the best trade we can make.

As far as I know, this should add full web support to the SDK. (#385)

### DIFF
--- a/lib/src/storage/core_store_sem_impl.dart
+++ b/lib/src/storage/core_store_sem_impl.dart
@@ -14,8 +14,16 @@ class CoreStoreSembastImp implements CoreStore {
       if (!parseIsWeb &&
           (Platform.isIOS || Platform.isAndroid || Platform.isMacOS))
         dbDirectory = (await getApplicationDocumentsDirectory()).path;
+      assert(() {
+        if (parseIsWeb) {
+          print(
+              'Warning: CoreStoreSembastImp of the Parse_Server_SDK does not encrypt the database on WEB.');
+        }
+        return true;
+      }());
       final String dbPath = path.join('$dbDirectory/parse', 'parse.db');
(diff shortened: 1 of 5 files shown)


phillwiggins:
Would this still allow encryption using Flutter but not Dart?

phillwiggins:
There are conflicts with this branch now.

nstrelow:
@phillwiggins 
The opposite right? Encryption as long as it is not running on web.
So Android, iOS and native dart probably.

@maaeps Maybe print a debug warning when running on web and using Sembast

maaeps:
All conflicts should be properly fixed now.

@nstrelow 
Great idea!

lib/src/storage/core_store_sem_impl.dart
```diff
@@ -14,8 +14,16 @@ class CoreStoreSembastImp implements CoreStore {
       if (!parseIsWeb &&
           (Platform.isIOS || Platform.isAndroid || Platform.isMacOS))
         dbDirectory = (await getApplicationDocumentsDirectory()).path;
+      assert(() {
+        if (parseIsWeb) {
+          print(
+              'Warning: CoreStoreSembastImp of the Parse_Server_SDK does not encrypt the database on WEB.');
```
nstrelow, line 16: Awesome :D

nstrelow (approved):

lib/src/storage/core_store_sem_impl.dart
maaeps: I found this on stackoverflow.
I have to remember this trick, it's super useful.

PR #441, opened 2020-09-10
Small changes and improvements

fischerscode:
Some small changes and improvements like changing the Parse Logo, adding "tests" and adding a default CoreStore for dart.


fischerscode:
Any idea why this fails? I have done nothing with travis before.

phillwiggins:
Don't worry, it runs the unit tests (Which there an extremely minimal amount, and are broken).

This package could really use some tests. I started it as an introduction to Dart but didn't TDD.

fischerscode:
Ok, should I remove 159d59aaac4530890fbe48e47c69bf72918c53fe again?

phillwiggins:
No it's fine. We should probably just update the tests that are there now to make Travis CI pass. There's only 2 I believe?

fischerscode:
There are currently almost no tests. They are more like, if I initialize the parse, is the application name set correctly.

PR #445, opened 2020-09-16
creat migration guide for version 1.0.28

fischerscode:
This is work in progress.
Due to the changed structure, creating a migration guide should be quite useful.


FNPCMDs (changes requested):
Hi! you misspelled some words, here is my review.

Change:
```
Strating with version 1.0.28, this repository is now seperated in a pure dart (parse_server_sdk) and a flutter package (parse_server_sdk_flutter).
This was done in order to provide a dart package for the parse-server, while keeping maintainance simple.
```
To:
```
Starting with version 1.0.28, this repository is now separated in a pure dart (parse_server_sdk) and a flutter package (parse_server_sdk_flutter).
This was done in order to provide a dart package for the parse-server, while keeping maintenance simple.
```

Also:
`As flutter web is now in beta, this SDK aims to be web compadible.`
To:
`As flutter web is now in beta, this SDK aims to be web compatible.`

And:
`But there are some parts completly different on web. For example the wep-app cant determine it's name, version or packagename.`
To:
`But there are some parts completly different on web. For example, the wep-app cant determine it's name, version or packagename.`

Furthermore, in Parse().initialize(), "appName", "appVersion", and "appPackageName" are optional arguments, so is not necessary define them with null.

fischerscode:
@FNPCMDs 
Thanks, normally my IDE shouts at me. Not sure why this didn't happen yesterday.

"appName", "appVersion", and "appPackageName" are optional arguments, but on web we have no chance to get this information. So we have to explicitly set them. Setting this values to null if we are not on web, results in them being automatically set. As far as I know, there is no way to set a optional argument only if a condition is true.

Issue #444, opened 2020-09-15
[Security] Storing password as plain text


BaranMichal25:
I just realised that the library is saving ParseUser data on most of successful [user operations](
https://github.com/parse-community/Parse-SDK-Flutter/blob/master/lib/src/objects/parse_user.dart#L430), which might seem handy, but the problem is that it is stored as [plain text](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/lib/src/objects/parse_base.dart#L181), **including password**, which is definitely a security issue.

I believe we should change library's behaviour here or at least make it more clear that the library is storing user passwords as plain text.
What do you think?

Or am I missing something? I know that encryption can be turned on in CoreStore library, but right now that would be suboptimal solution, because all data would be automatically encrypted.

RodrigoSMarques:
Hi @BaranMichal25 

Using Sembast, encryption is enabled by default, and you can enter an encryption password.
```dart
     ...
      coreStore: await CoreStoreSembastImp.getInstance(password: 'password'),
     ...
```

BaranMichal25:
Yes, but then all data will be encrypted, right?
In which case, also loading all other classes and their data would be needlessly slow

RodrigoSMarques:
Yes, all data will be encrypted.

Is it really slow? How much data will you store locally?
Sembast is a pure dart database and very fast in performance tests.

fischerscode:
CoreStoreSembastImp is always encrypted (except on web), the password has a [default value](https://github.com/parse-community/Parse-SDK-Flutter/blob/2088bf1b07594a080b91d5002eda0515dae63568/lib/src/storage/core_store_sp_impl.dart#L9).

I think, the password should not be stored in any database neither in clear text, nor encrypted. Only storing a hashed password is acceptable.

When inspecting my saved user object (`ParseUser.currentUser()`), it indeed still contains the password.

I have three solutions to solve this problem:
1. provide the password when calling login()
2. storing the password in a field (instead of `_objectData`) (should be the easiest to implement)
3. removing the password at the `saveInStorage` call

I think the second one is the best. I will create a branch to test this change, soon.

fischerscode:
I have implemented option two in [this branch](https://github.com/fischerscode/Parse-SDK-Flutter/tree/issue/%23444).
Testing is highly appreciated as I do not know which side effects this change has.

fischerscode:
[...]

PR #446, opened 2020-09-16
ParseUser: store password in field

fischerscode:
Potential fix for #444.
But some more testing is required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,12 @@ dart:
 
 install:
   - git clone https://github.com/flutter/flutter.git -b stable --depth 1
-  - export PATH=./flutter/bin:$PATH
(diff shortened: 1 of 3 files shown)


fischerscode:
@phillwiggins 
Can you think of potential problems due to this change?
Apart from not having the password copied when doing something like `ParseUser user2 = ParseUser(null, null, null).fromJson(user.toJson());`, everything should work like before, right?

phillwiggins:
Hmmmm, I really don't know to be honest. I suppose the only way to work that out is to quickly scan through the docs and understand what requests are being made with the password inside the request. I would assume it's only login etc? This could be a viable solution?

fischerscode:
I did a quick scan and I think it should everything work.
I will merge this as soon as the tests finish.

PR #474, opened 2020-10-09
Add disclaimer to not use Masterkey on client side

### LINKED ISSUE #473: Masterkey is not required and should not be available on client side
I just learned today, that the Masterkey is only required for changing ACLs. I changed the key and my app was still working. Had a project for 6 months with Masterkey defined in the client app. Luckily not a production app.

Parse Server Readme: https://github.com/parse-community/parse-server#basic-options

> The master key to use for overriding ACL security. You can use any arbitrary string. Keep it secret! For migrated apps, this should match your hosted Parse app.

Parse Client JS SDK Docs: https://docs.parseplatform.org/js/guide/

> ⚠️ If the Masterkey needs to be provided, use the following. Please note that the master key should only be used in safe environments and never on client side ‼️

Maybe we should add a similar strong disclaimer to our Readme?

nstrelow:
Fixes #473

### DIFF
--- a/packages/dart/README.md
+++ b/packages/dart/README.md
@@ -57,7 +57,7 @@ It's possible to add other parameters to work with your instance of Parse Server
         securityContext: securityContext, // Again, required for some setups
 	coreStore: CoreStoreMemoryImp()); // Non persistent mode (default): Sdk will store everything in memmore instead of using Sembast as an internal DB.
 ```
-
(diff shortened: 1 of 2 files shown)


fischerscode:
I think it's a good idea to add this to the README.
Thanks!

nstrelow:
Danke dir!

fischerscode:
Ganz unerwartet auf github mal was deutsches zu lesen :)

PR #499, opened 2020-10-29
Fix CoreStoreSembastImp / Enhancements LiveQuery / New Query Method

RodrigoSMarques:
- Fix CoreStoreSembastImp with no default passwod causes error in Flutter Package
- Makes LiveQuery a class that is not Singleton. This allows you to use more than one LiveQuery at the same time in the app
- Fix ParseACL with no default value in setReadAccess and setWriteAccess
- Add method excludeKeys: Exclude specific fields from the returned query

### DIFF
--- a/packages/dart/CHANGELOG.md
+++ b/packages/dart/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.0.1
+Fixed network exceptions. [#482](https://github.com/parse-community/Parse-SDK-Flutter/pull/482)
+
 ## 2.0.0
 ##### Warning: This release contains breaking changes. If you are using flutter you should migrate using *[this](https://github.com/parse-community/Parse-SDK-Flutter/blob/release/2.0.0/docs/migrate-2-0-0.md)* guide.
 
--- a/packages/dart/README.md
+++ b/packages/dart/README.md
@@ -20,7 +20,7 @@ This is a work in progress and we are consistently updating it. Please let us kn
 To install, either add to your pubspec.yaml
 ```yml
 dependencies:  
-    parse_server_sdk: ^2.0.0
+    parse_server_sdk: ^2.0.1
 ```
 or clone this repository and add to your project. As this is an early development with multiple contributors, it is probably best to download/clone and keep updating as an when a new feature is added.
 
--- a/packages/dart/lib/src/base/parse_constants.dart
+++ b/packages/dart/lib/src/base/parse_constants.dart
@@ -1,7 +1,7 @@
 part of flutter_parse_sdk;
 
 // Library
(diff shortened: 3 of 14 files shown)


fischerscode:
> Makes LiveQuery a class that is not Singleton. This allows you to use more than one LiveQuery at the same time in the app

Could you please explain this to me? From my knowledge, it has been possible to have multiple `LiveQueries` at the same time. Sure, they used the same `LiveQueryClient` but isn't this a performance advantage, as there are less connections to the server?

RodrigoSMarques:
@fischerscode 

I checked LiveQuery on Android and it's not Singleton. I have a native app (java) with more than two LiveQuery.

Let's take for example Whatsapp, where you have a list of conversations that needs to be updated when a message arrives and a list of messages that needs to be updated in the chat.

My current case on Flutter, I have a screen with two sessions. Each session needs to be updated live by different LiveQuery, due to the QueryBuilder criteria.

I've been testing it for a few days and I haven't identified any performance issues. When I leave the screen, I disconnect LiveQuery and release that session on the server.

I had a problem with the current implementation. As a SingleTon, the session is active for the entire life of the app.

When disconnecting without passing the userInitialized parameter, LiveQuery was doing reconnect and creating multiple connections, which reached the point of bursting my server's memory.

And this was multiple times. It took me a long time to realize that a session of the app that was being tested on my simulator was that it was recreating multiple sessions on the server.

fischerscode:
@RodrigoSMarques 
Maybe I misunderstand your problem.

> When I create a LiveQuery, I cannot have another one with different criteria.

I am using ParseLiveList (which uses LiveQuery) all over my app and having two visible lists with different queries works as expected. So I am pretty sure this works:
> Let's take for example Whatsapp, where you have a list of conversations that needs to be updated when a message arrives and a list of messages that needs to be updated in the chat.

fischerscode:
@RodrigoSMarques 
I've created an [example](https://github.com/fischerscode/Parse-SDK-Flutter/tree/example_multiple_livequeries/packages/dart/example_multiple_livequeries) how I use multiple `LiveQueries`.

RodrigoSMarques:
@fischerscode 

I have to confess that I rushed into this PR.

After your comment I redid the tests on another project that I just created. I analyzed the code calmly and managed to use LiveQuery with two different Query.

LiveQuery is just a client and in that sense it makes sense to be a Singleton, just like ParseHTTPClient.

Now I'm going to check what's wrong with my project. Because it didn't work there.

I will cancel the PR and keep only the other changes.

fischerscode:
@RodrigoSMarques 
I wish you good luck finding the issue.
For me, non working LiveQueries have often been caused by forgetting to tell the server that the new table should be "livequeryable".

PR #503, opened 2020-11-05
Fix CoreStoreSembastImp / New Query Method / Update Dependencies version

RodrigoSMarques:
- Fix CoreStoreSembastImp with no default passwod causes error in Flutter Package
- Fix ParseACL with no default value in setReadAccess and setWriteAccess
- Add method excludeKeys: Exclude specific fields from the returned query
- Update Dependencies version

### DIFF
--- a/packages/dart/lib/src/network/parse_query.dart
+++ b/packages/dart/lib/src/network/parse_query.dart
@@ -86,6 +86,13 @@ class QueryBuilder<T extends ParseObject> {
     limiters['keys'] = concatenateArray(keys);
   }
 
+  ///Exclude specific fields from the returned query
+  ///
+  /// [String] keys not will return the columns of a result you want the data for
+  void excludeKeys(List<String> keys) {
+    limiters['excludeKeys'] = concatenateArray(keys);
+  }
+
   /// Includes other ParseObjects stored as a Pointer
   void includeObject(List<String> objectTypes) {
     limiters['include'] = concatenateArray(objectTypes);
--- a/packages/flutter/lib/parse_server_sdk.dart
+++ b/packages/flutter/lib/parse_server_sdk.dart
@@ -143,7 +143,7 @@ class CoreStoreSembastImp implements sdk.CoreStoreSembastImp {
   static sdk.CoreStoreSembastImp _sembastImp;
 
   static Future<sdk.CoreStore> getInstance(
-      {DatabaseFactory factory, String password}) async {
+      {DatabaseFactory factory, String password = 'flutter_sdk'}) async {
     if (_sembastImp == null) {
       String dbDirectory = '';
       if (!sdk.parseIsWeb &&
--- a/packages/dart/CHANGELOG.md
+++ b/packages/dart/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.0.1
+Fixed network exceptions. [#482](https://github.com/parse-community/Parse-SDK-Flutter/pull/482)
+
 ## 2.0.0
 ##### Warning: This release contains breaking changes. If you are using flutter you should migrate using *[this](https://github.com/parse-community/Parse-SDK-Flutter/blob/release/2.0.0/docs/migrate-2-0-0.md)* guide.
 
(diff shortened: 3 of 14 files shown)


packages/dart/lib/src/network/parse_query.dart
```diff
@@ -86,6 +86,13 @@ class QueryBuilder<T extends ParseObject> {
     limiters['keys'] = concatenateArray(keys);
   }
 
+  ///Exclude specific fields from the returned query
+  ///
+  /// [String] keys not will return the columns of a result you want the data for
+  void excludeKeys(List<String> keys) {
+    limiters['excludeKeys'] = concatenateArray(keys);
+  }
+
```
fischerscode, line 95: Is it like the opposite of `keysToReturn(List<String> keys)`?

packages/flutter/lib/parse_server_sdk.dart
```diff
@@ -143,7 +143,7 @@ class CoreStoreSembastImp implements sdk.CoreStoreSembastImp {
   static sdk.CoreStoreSembastImp _sembastImp;
 
   static Future<sdk.CoreStore> getInstance(
-      {DatabaseFactory factory, String password}) async {
+      {DatabaseFactory factory, String password = 'flutter_sdk'}) async {
```
fischerscode, line 146: I think this will avoid some frustration during development. 👍 
Do you think it would be a good idea to print a warning to the developer if the default password is used? (Only in debug mode using the assert-trick)

fischerscode (approved):
👍

packages/flutter/lib/parse_server_sdk.dart
fischerscode: I think it fits this PR.
Could you please add `?? 'flutter_sdk'` behind `masterKey` in this [line](https://github.com/parse-community/Parse-SDK-Flutter/blob/30ba45c86a17054b602c23ac3b718f2274db5b32/packages/flutter/lib/parse_server_sdk.dart#L89)?
(If you are busy, I can do it as well.)

packages/dart/lib/src/network/parse_query.dart
RodrigoSMarques: Yes. The documentation is [here](https://docs.parseplatform.org/rest/guide/#query-constraints).
Maybe you just want to remove one field from the result and this constraint makes it easier.

packages/flutter/lib/parse_server_sdk.dart
RodrigoSMarques: The parse_server_sdk package has a default password. When dividing the package into Dart / Flutter the password was not defined in the flutter and this causes an error in the initialization.
Yes. An alert might be interesting

fischerscode: When I split the package, I did not realize, the value `null` would be 'forwarded'. As a result, the default password in the dart file is not used. My bad!

Something like this?
```dart
assert(() {
  if (password == 'flutter_sdk') {
    print('Warning: CoreStoreSembastImp uses the default password. Specify a custom one for increased security.');
  }
  return true;
}());
```
(Should be changed in packages/dart/lib/storage/core_store_sem_impl.dart)

packages/dart/lib/src/network/parse_query.dart
fischerscode: Great! 👍

packages/flutter/lib/parse_server_sdk.dart
RodrigoSMarques:
```dart

assert(() {
  if (password == 'flutter_sdk') {
   print('***********************************************************************************************************');
   print('Warning: CoreStoreSembastImp uses the default password. Specify a custom one for increased security.');
   print('***********************************************************************************************************');
  }
  return true;
}());

````

fischerscode: 👍 A little more visible.

fischerscode:
@RodrigoSMarques 
I am kind a surprised that this PR is still pending. I assume you are waiting for me to merge it, right?

fischerscode:
Just noticed, this PR would merge into master. Please change it to development.

RodrigoSMarques:
Hi @fischerscode 
I was inattentive when sending to the master branch. I will send a new PR to branch development

PR #508, opened 2020-11-15
add typings to getRelation() function

arbue:
Adding typings to ParseObject's getRelation method.

PR #513, opened 2020-11-18
Correctly encode value in addConfig() function

formatCvt:
Correctly encode value in ParseConfig.addConfig() function (fix hardcoding to string and allow to use boolean, number, object and so on)


fischerscode:
(The tests should fail for the same reason they failed in #508. So it should be fine.)

PR #507, opened 2020-11-13
ParseUser.loginWith(...) username, password, email

Danaru87:
Add username, password and email as optional params for ParseUser.loginWith(...). Related: #506

Issue #506, opened 2020-11-13
Specify email address in LoginWith 


Chralu:
Hi,

when login with an third party auh service for the first time, I would like to set email without triggering email verification.
Indeed, `email` should already be verified by upstream third auth service 

We could modify the `loginWith` method :
```
ParseUser.loginWith(String provider, AuthData authData, {String email});
```

That way, `email` is set during the `POST /parse/users`, and email verification is not triggered.

Any thoughts about that ?

Danaru87:
# In Parse Server REST GUIDE
there are two methods to register user with third party provider:

## POST /parse/user
Sending only authData in body request, this will create a new user with random username, random password. Other user's properties are ignored.

## POST /parse/user
Sendind json body containing user's properties AND authData property, this will create a new user linked to third party provider, setting all properties set in json body into user's record.

### Example:

**Register only with AutData:**

**Request**
```
curl --request POST \
  --url https://{parseServerUrl}/parse/users \
  --header 'Content-Type: application/json' \
  --header 'X-Parse-Application-Id: {appId}' \
  --data '{"authData":{"provider":{"access_token":"accessToken","id":"id","id_token":"id_token"}}}'
```
**Result**
created user as json:
```json
{
"objectId":"objectId",
"username":"generated",
"password":"*********",
"email": "undefined",
"authData":{"authData":{"provider":{"access_token":"accessToken","id":"id","id_token":"id_token"}}},
"emailVerified": "undefined",
"updatedAt": "....",
"createdAt": ".....",
"ACL": "...."
}
```

**Register AutData and User properties:**

**Request**
```curl
curl --request POST \
  --url https://{parseServerUrl}/parse/users \
  --header 'Content-Type: application/json' \
  --header 'X-Parse-Application-Id: {appId}' \
  --data '{
"username":"username",
"password":"password",
	"email": "email@email.io",
	"authData":{"google": {"access_token":"accessToken", "id":"id", "id_token": "idToken"}}
}'
```
**Result**
created user as json:
```json
{
"objectId":"objectId",
"username":"username",
"password":"*********",
"email": "email@email.io",
"authData":{"google": {"access_token":"accessToken", "id":"id", "id_token": "idToken"}},
"emailVerified": "undefined",
"updatedAt": "....",
"createdAt": ".....",
"ACL": "...."
}
```

# In Parse SDK 

In dart/flutter SDK, only **Register with AutData** is possible.
[...]

PR #514, opened 2020-11-19
ParseUser.loginWith(...) username, password, email

Danaru87:
Replace #507 

Add username, password and email as optional params for ParseUser.loginWith(...). Related: #506


fischerscode:
Great 👍

PR #518, opened 2020-11-20
Update parse_live_list.dart

pastordee:
added Build Animated Grid to ParseLiveListWidget


fischerscode:
Interesting! Thanks for your PR 👍

Do you think it would be a good idea to put this change in a separated `ParseLiveGridWidget`?
I think by doing so, more developers would find this functionality and it might be easier to maintain.

pastordee:
@fischerscode 
Yes that will be nice you can give it a go.

fischerscode:
@pastordee 
I've added a commit that fixed a bug when an item was added/removed.

Thanks for you contribution, I am merging the changes now.

Issue #552, opened 2021-01-17
Widget for ParseLiveList when query is empty


pastordee:
Can we have a widget for when the query is empty so you can show  message to the client like listLoadingElement.

PR #563, opened 2021-02-11
Empty Query Widget ParseLiveList

pastordee:
When query is empty you can show a widget like (listLoadingElement).
#552 (comments)


RodrigoSMarques:
Hi @pastordee .
Thanks for your PR.

When calling setState within `query.count()`, is the `setState` that build the ParseLiveList called? Is this code working correctly?

You are making a query in Parse to count the number of objects to check whether the query is empty or not. Two requests to Parse are required to set up the ParseLiveList. This can be a problem for some applications.

PR #585, opened 2021-03-20
Fix LiveQuery: wont add element to empty list

fischerscode:
Version 2.1.0 has a bug, where new rows won't be added to an empty list.

This occurred, as the animated list was not build in case of an empty list. Resulting in `_animatedListKey.currentState` being null. By using a stack to view both widgets, this should be fixed without any side effects.

The `queryEmptyElement` is now also animated.

### DIFF
--- a/packages/dart/lib/parse_server_sdk.dart
+++ b/packages/dart/lib/parse_server_sdk.dart
@@ -6,6 +6,7 @@ import 'dart:io';
 import 'dart:math';
 import 'dart:typed_data';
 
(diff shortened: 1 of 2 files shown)


RodrigoSMarques:
Hello @fischerscode .

Merge the PR on the nullsafety branch, which will be the next one that we will make available.
There was conflict. Try to check them out.

fischerscode:
Closing this, I've recreated the PR: #608

PR #606, opened 2021-04-22
Added UpdateEvent on ParseLiveList

gcostaapps:
It allows users who are listening to the stream in LiveList to know when a data was updated.


fischerscode:
Thanks for the PR!

(I think the failing test is not related to this change, as the test on the latest commit of the target branch failed, too.)

PR #608, opened 2021-04-23
Fix LiveQuery: wont add element to empty list

fischerscode:
original PR: #585 

> Version 2.1.0 has a bug, where new rows won't be added to an empty list.
> 
> This occurred, as the animated list was not build in case of an empty list. Resulting in _animatedListKey.currentState being null. By using a stack to view both widgets, this should be fixed without any side effects.
> 
> The queryEmptyElement is now also animated.

I've recreated this PR, as the original has been quite old and there have been some changes in the files.


fischerscode:
Merging this now, as it seems to fix the issue.

Issue #610, opened 2021-04-24
additional query methods


fischerscode:
I propose the addition of `Future<T>? first();` and `Future<List<T>> find();` to the `QueryBuilder` as the add a easy to use alternative to the `Future<ParseResponse> query();` method.

The advantages (at least I think they are) over the `Future<ParseResponse> query();` method are the following:
- Although within the SDK Type errors are still possible, the developer gets the option of an type "save" query method.
- In some cases an empty list or a null result is easier to handle then a `ParseResponse` containing an error and a null result.

Every developer would have the option to chose the query method, that suits the individual problem and the personal coding style best.

Personally, I've added methods like this using an extension, but I think adding methods like this to the SDK would be a nice addition. 

I've already created #611 containing those new methods.

RodrigoSMarques:
@fischerscode
Good proposal. It would be important to just document it in README.MD and you can merge it yourself.

PR #611, opened 2021-04-24
Additional query methods

fischerscode:
Please see #610 for more information.


pastordee:
@fischerscode 
Hi is it possible to get ParseLiveListWidget to reload itself when it connects back to the Internet

PR #618, opened 2021-04-30
Fix error of live query when an object is added/deleted with a null pointer that is on includeObjects

gcostaapps:


RodrigoSMarques:
Thanks for your PR.

Issue #603, opened 2021-04-20
ParseObject.getRelation returns incomplete ParseRelation


Mabenan:
When you exectute the following code you get an exception

`
Receipt objectId = Receipt().getObject("objectId");
elementAt.getRelation<Price>("Prices").getQuery();
`
the reason for this is in getRelation the relation is instantiated without setting the attribute _targetClass
![image](https://user-images.githubusercontent.com/4959465/115402328-9db1e500-a1eb-11eb-9f73-1b21227b3540.png)
which is needed in getQuery()
![image](https://user-images.githubusercontent.com/4959465/115402422-b15d4b80-a1eb-11eb-9340-a533cca6b963.png)
The attribute is only set when you execute fromJson or if you  add/remove something from the relation.

Possible Solution would be to add _targetClass in the constructor and/or instance the object when T is set.

RodrigoSMarques:
My sample code:

```dart
    final parseRelation = resultTestAPI.getRelation('productRelation');
    parseRelation.add(ParseObject("Product")..set("objectId", "XGadzYxnac"));

    print('********************************************************');
    print('Save relation - Add Object');
    print('********************************************************');
    parseObject.set<ParseRelation>('productRelation', parseRelation);

    apiResponse = await parseObject.save();

    print(apiResponse.results.toString());

    final relationResponse = await (parseRelation.getQuery()).query();
    if (relationResponse.success) {
      print('sucesso relation ${relationResponse.results.length}');
    }
```

Mabenan:
Yeah but I just want to read, not change the related items.

RodrigoSMarques:
It's in the example, just ignore the save part

```dart
   final parseRelation = resultTestAPI.getRelation('productRelation');
   final relationResponse = await (parseRelation.getQuery()).query();
    if (relationResponse.success) {
      print('sucesso relation ${relationResponse.results.length}');
    }
```

Mabenan:
Yeah but the call getQuery() fails because _targetClass is initial. I currently use whereRelatedTo as a workaround but it would be better if I can use getQuery

RodrigoSMarques:
Strange. I ran my sample code now and an error occurred.
I made the correction in the migration from version 2x to 3x and it was working.
There may have been an error in the merge.
I will evaluate.

Tehforsch:
[...]

PR #622, opened 2021-05-04
Added Test and Fix for Issue #603

Mabenan:
Signed-off-by: Mabenan <doener123@googlemail.com>

This should fix the Issue #603.
Sadly Dart doesn't provide a good way to instantiate a generic otherwise it would be possible to just use the generic.


RodrigoSMarques:
Fixed in version 3.1.0. Another solution was implemented, without the need to pass parameters

PR #625, opened 2021-05-09
Fix TypeError thrown when parsing objects with non-string createdAt or updatedAt values in maps

Molkars:


Molkars:
Pull Request to wrong branch

PR #626, opened 2021-05-09
Prevent a TypeError from being thrown while parsing an object from a map whose createdAt or updatedAt properties are not Strings

Molkars:


RodrigoSMarques:
Thanks for your PR.

PR #637, opened 2021-06-04
Expose additionalHeaders field using a getter and setter on the paren…

### LINKED ISSUE #635: Can't configure additionalHeaders in ParseHTTPClient anymore.
I'm unable to set additionalHeaders on the ParseHTTPClient class anymore.
Previously this field was publicly accessible, but now it has been moved to a private class which is used in a private field.
This bug prevents me from migrating to the current version of the library.

**Affected Versions**
\>=2.1.0

**Caused by**
Added ParseClient to abstract dio and http (#495)
Commit e37236fa15c90e54e36ee19069c01d8ea0ce9c10

**Proposed Solution**
Expose the additionalHeaders field from the _ParseHTTPClient class outside of the ParseHTTPClient class using a getter/setter.

I will create a PR for this and add it to this issue.

Moomba42:
Fixes #635 on the nullsafety branch

### DIFF
--- a/packages/dart/lib/src/network/parse_http_client.dart
+++ b/packages/dart/lib/src/network/parse_http_client.dart
@@ -16,6 +16,11 @@ class ParseHTTPClient extends ParseClient {
 
   late _ParseHTTPClient _client;
 
+  Map<String, String>? get additionalHeaders => _client.additionalHeaders;
+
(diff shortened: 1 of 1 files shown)


RodrigoSMarques:
Thanks for your PR.

PR #638, opened 2021-06-06
:bug: containsKey on wrong key

bernhardfrenking:
The containsKey check is on the wrong key (keyVarObjectId) instead of (keyParamSessionToken) - therefore it fails on null safety, if the sessionToken is not provided in responseData (for example in case of signUp)


RodrigoSMarques:
Thanks for your PR.

PR #720, opened 2022-04-30
ci: migrate from TravisCI to GitHub Actions

fischerscode:

### Issue Description
This PR migrates the tests to GitHub Actions.

Related issue: #721 

### Approach
I've decided to
- test both packages independently.
- run the tests on ubuntu, macos and windows. I don't know if it's really needed for this project, but the [example](https://github.com/dart-lang/setup-dart#matrix-testing-example) does it...
- run the tests on the stable,beta,dev branch. (dev only for the dart package)
- test pushes to any branch and PRs.
- override the `parse_server_sdk` dependency in the flutter package to the relative path.

Feedback/critic is obviously welcome.

### TODOs before merging

Maybes:
- [ ] Fix the tests
- [ ] Add code coverage reports

Since testing this is almost impossible, it's untested.


fischerscode:
@mtrezza That was fast :D

mtrezza:
I'm just trying to catch up with you ;-)

fischerscode:
I think the tests don't run, since github uses the master workflows for new PRs.

mtrezza:
They should run, I think you need to change something in the workflow file, like:

```yaml
name: ci
on:
  push:
    branches: [ master, development ]
  pull_request:
    branches:
      - '**'
```

fischerscode:
> We do not have any branches named like that?

Those are the names of the flutter/dart project channels.

mtrezza:
How do these channels correspond to our branches?

fischerscode:
> How do these channels correspond to our branches? In other words, from which branch do we release to which channel?

There is no connection between the channels / branches of the flutter project and our branches.
I just test the code with different flutter versions to make sure it also works with the current flutter beta. (Same applies for dart)

mtrezza:
So have we published anything in the `dev` and `beta` channels in the past, or only in the `stable` channel?

fischerscode:
I think branches for `pull_request` are just optional. The idea might be that it would be bad if a user could steal secrets by changing the `pull_request` workflows.

mtrezza:
Maybe you were right and this has to be merged into master first to get triggered on PRs.

fischerscode:
I'll test it in a dummy repo.

fischerscode:
push works as expected: https://github.com/fischerscode/Parse-SDK-Flutter/actions/runs/2248353523
still testing pull_request

.github/workflows/ci.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Run tests
+        run: (cd packages/dart && dart test)
+
+  test-flutter:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        channel: [stable, beta]
```
mtrezza, line 50: Why is there no `dev` channel but there is for dart?

mtrezza (changes requested):
Just some minor formatting to keep the structure the same as with our other workflows.

.github/workflows/ci.yml
fischerscode: grafik
On the top, it has not been listed as a valid option, but as I've noticed now, it seems to be (bellow). I'll add it.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
+name: ci
+on:
+  push:
+    branches:
+      - "master"
+      - "development"
+    paths-ignore:
```
mtrezza, line 7: That is a really good catch, we should add this to our existing workflows in the other repos, thanks!

mtrezza (approved):
Looks good! Let me know when you think this is ready for merge.

fischerscode:
With the dummy repo research in mind, think it's ready.

Expected behavior: Tests should fail until #722 is fixed.

mtrezza:
So what is the purpose of running a test per channel? The test runs the same code for each channel, and the channel is only relevant for publishing, right?

fischerscode:
Stuff might break in the beta/dev channel and by testing it before it's stable, we could investigate if it's our fault or a bug in the flutter/dart project. But to be honest, that's quite unlikely, we could also consider testing only on stable and linux.

fischerscode:
Just noticed the 
grafik
thing is now outdated since it uses travis.

mtrezza:
I fixed the badge in https://github.com/parse-community/Parse-SDK-Flutter/pull/723

mtrezza:
> Stuff might break in the beta/dev channel 

Why? I thought it was only a release channel. What's different in each of these channels?

fischerscode:
Flutter has [multiple release channels](https://github.com/flutter/flutter/wiki/Flutter-build-release-channels), [the same applies for dart](https://dart.dev/get-dart#release-channels), this has nothing to do with pub.dev.

We publish to pub.dev, there we can release stable versions and prereleases (semantic version with a "-" and some text).

I think it might be good to check if our code works with "any" of the latest flutter/dart versions. There for testing it with the stable and beta (and dev) version of flutter and dart.

However I'm not sure if testing on windows/macos is really necessary, since our code is (currently) not platform specific and therefore there should be no difference when running it.

Of course it is valid to discus if testing on an other channel then the stable one is even necessary (esp. dev).

By the way:
> There used to be a dev branch as well (and you may still see it in our tooling). We no longer update this branch and our tooling will be updated in due course to no longer list it (if it hasn't already). (https://github.com/flutter/flutter/wiki/Flutter-build-release-channels#which-channel-should-i-use)

We should not test the flutter package on the dev channel.

mtrezza:
> There for testing it with the stable and beta (and dev) version of flutter and dart.

Got it, this does not refer to our release type, but to the flutter / dart stable / beta / dev releases, correct?

fischerscode:
Correct.

mtrezza:
In that case I don't see any reason to test against the `dev` channel, maybe not even the `beta` channel. These CI checks may fail because these are not stable releases, which has nothing to do with our code, therefore it probably shouldn't be in our code's CI.

fischerscode:
Yesterday I've been more on the "make sure it also works with upcoming versions" side, but seeing the large list of tests, I tend to your side.
grafik

MongoDB etc. is a good point, too.

fischerscode:
What do you think about testing on all operating systems? I tend to leave it to linux until we have some code that is really platform dependent.

mtrezza:
> "make sure it also works with upcoming versions"

Yeah, I get your point, it would be nice to see whether something "will likely work" with an upcoming stable release. The counter argument are probably:
- All other Parse Org repos only have "must pass" CI checks; a failing "optional" CI check may be difficult to interpret for reviewers, so I would leave this all out.
- Even if an "optional" check fails, since it's not a stable release it probably also doesn't make sense to adapt the PR in any way, as the prerelease code can change at any time.

> testing on all operating systems?

I am not familiar with how easy one can introduce platform specific code in a PR. If that is just a condition like `if platform_macos { ... }` then I would test against all platforms; if introducing platform specific code is a more complex feature on it's own that will require its own PR, then I would probably leave it to linux for now.

fischerscode:
> how easy one can introduce platform specific code in a PR

Should not happen at this package, the idea is, that the code gets compiled differently depending on the platform, but I think it's best to just test on linux for now.

> it would be nice to see whether something "will likely work" with an upcoming release

Does github support optional checks?

It might be best to keep it simple for now, but I would keep the matrix setup, that way it's simple to add an other channel/platform.

mtrezza:
> Does github support optional checks?

It does, but I agree with you keeping it simple for now.

Issue #725, opened 2022-04-30
Fixing the failing tests


fischerscode:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

~~- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). ~~

### Issue Description
Currently the tests are failing.

We have to:
- [x] #722 
- [x] #728 
- [x] #729

### Steps to reproduce
n/a

### Actual Outcome
n/a

### Expected Outcome
n/a

### Environment
GitHub Ci

### Logs
n/a

mtrezza:
Could you please open an issue per PR? We have this requirement because sometimes there are discussions around individual changes, also maybe at a later point in time, and we want to keep them separated by issue.

fischerscode:
I'll link the issues in the list.

mtrezza:
Got it, you opened this a sort of roadmap for issues to fix, not to reference PRs.

fischerscode:
Well, you did not...
At first, I thought about linking the PRs to this issue and decided to use this one as a roadmap after I read your comment.

mtrezza:
You are on a roll! Amazing how many things we got done in the last 24h.

mtrezza:
Closing as all tests pass.

PR #726, opened 2022-04-30
style: format code to pass lint CI check

fischerscode:

### Issue Description
This PR formats the code as wanted by #725.

Related issue: #725 

### Approach
Just dart/flutter format

### DIFF
--- a/packages/dart/lib/src/objects/parse_file_base.dart
+++ b/packages/dart/lib/src/objects/parse_file_base.dart
@@ -16,9 +16,7 @@ abstract class ParseFileBase extends ParseObject {
             client: client) {
     _path = '/files/$name';
     this.name = name;
-    if(url != null)
-      this.url = url;
-
+    if (url != null) this.url = url;
   }
 
   String get name => super.get<String>(keyVarName)!;
--- a/packages/dart/lib/parse_server_sdk.dart
+++ b/packages/dart/lib/parse_server_sdk.dart
@@ -139,7 +139,7 @@ class Parse {
 
   bool hasParseBeenInitialized() => _hasBeenInitialized;
 
-  Future<ParseResponse > healthCheck(
+  Future<ParseResponse> healthCheck(
       {bool? debug, ParseClient? client, bool? sendSessionIdByDefault}) async {
     final bool _debug = isDebugEnabled(objectLevelDebug: debug);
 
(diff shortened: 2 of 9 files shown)


mtrezza:
Which CI check currently checks the formatting?

mtrezza:
If this is merely a code-refactor, then the commit type is `refactor:`.

pastordee:
Hey thank you guys for the hard work I’ve been watching you guys for the past 24 hours I don’t think you have slept well done and God bless.

fischerscode:
Lint succeed, so this the test seems to work.

mtrezza:
Thanks @pastordee, want to join in? 😁

mtrezza:
> Lint succeed, so this the test seems to work.

So we can merge this PR? Or is there more to do?

fischerscode:
I think we can merge.

packages/dart/lib/src/objects/parse_file_base.dart
```diff
@@ -16,9 +16,7 @@ abstract class ParseFileBase extends ParseObject {
             client: client) {
     _path = '/files/$name';
     this.name = name;
-    if(url != null)
-      this.url = url;
-
+    if (url != null) this.url = url;
```
mtrezza, line 19: I'm not really a fan a unscoped if statements; but I guess that's for another PR...

fischerscode:
grafik
Why suddenly two? Is it because of "style:"?

mtrezza (approved):
Looks good!

packages/dart/lib/src/objects/parse_file_base.dart
fischerscode: I think there is a way to change the flutter style... maybe analysis_options.yaml?
I can research how this can be changed.

mtrezza:
> Why suddenly two? Is it because of "style:"?

I changed the settings to align with our other repo settings. Also, we have added the secrets to the workflow, so the stakes are high now and we will probably need to restrict merging PRs to Parse Org maintainers to mitigate the risk of the packages getting highjacked through a malicious PR.

packages/dart/lib/src/objects/parse_file_base.dart
fischerscode: I would leave it like this for now and create an issue to revisit the lint settings.

mtrezza: I think that would be good practice; we don't allow that in other repos because it's easy to misread such lines for developers.

Issue #721, opened 2022-04-30
Migrate from TravisCI to GitHub actions


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Current Limitation
We are migrating all repositories to GitHub Actions across Parse Platform; the CI of this repository is still using TravisCI.

### Feature / Enhancement Description
Use GitHub Actions.

### Example Use Case
n/a

### Alternatives / Workarounds
n/a

mtrezza:
As I said, I'm trying to catch up with you ;-)

mtrezza:
Reopened as we need to do some optimization, see https://github.com/parse-community/Parse-SDK-Flutter/pull/727

PR #727, opened 2022-04-30
ci: optimize CI workflow

mtrezza:

### Issue Description
- move lint into separate job
- use caching for flutter download
- test only against stable flutter / dart versions (not beta, dev)

Related issue: https://github.com/parse-community/Parse-SDK-Flutter/issues/721

### Approach
Fixed issues mentioned above.

### TODOs before merging
n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
+      - name: Install dependencies
+        run: dart pub get --directory packages/dart
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/dart
   check-code-analysis-flutter:
@@ -46,6 +48,13 @@ jobs:
         uses: subosito/flutter-action@v2
         with:
           cache: true
+      - name: Install dependencies
+        run: |
+          (cd packages/dart && flutter pub get)
+          cd packages/flutter
+          flutter pub remove parse_server_sdk
+          flutter pub add parse_server_sdk --path ../dart
+          flutter pub get
       - name: Analyze code
         run: flutter analyze packages/flutter
   check-code-analysis-dart:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+      fail-fast: false
     name: Test flutter (${{ matrix.os }})
     steps:
       - name: Checkout code
@@ -87,6 +88,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+      fail-fast: false
     name: Test dart (${{ matrix.os }})
     steps:
       - name: Checkout code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,30 @@ jobs:
           cache: true
       - name: Lint
         run: flutter format --output=none --set-exit-if-changed packages/flutter
+  check-code-analysis-flutter:
+    name: Code analysis (flutter)
+    timeout-minutes: 5
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup flutter
+        uses: subosito/flutter-action@v2
+        with:
+          cache: true
+      - name: Analyze code
+        run: flutter analyze packages/flutter
+  check-code-analysis-dart:
+    name: Code analysis (dart)
+    timeout-minutes: 5
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup dart
+        uses: dart-lang/setup-dart@v1
+      - name: Analyze code
+        run: dart analyze packages/dart
   check-dart:
     runs-on: ${{ matrix.os }}
     strategy:
@@ -41,8 +65,6 @@ jobs:
         run: dart pub get --directory packages/dart
       - name: Run build_runner
         run: (cd packages/dart && dart run build_runner build --delete-conflicting-outputs)
-      - name: Analyze project source
-        run: dart analyze packages/dart
       - name: Run tests
         run: (cd packages/dart && dart test)
   check-flutter:
@@ -58,6 +80,7 @@ jobs:
         uses: subosito/flutter-action@v2
         with:
           channel: ${{ matrix.channel }}
+          cache: true
       - name: Install dependencies
         run: |
           (cd packages/dart && flutter pub get)
@@ -65,7 +88,5 @@ jobs:
           flutter pub remove parse_server_sdk
           flutter pub add parse_server_sdk --path ../dart
           flutter pub get
-      - name: Analyze project source
-        run: flutter analyze packages/flutter
       - name: Run tests
         run: (cd packages/flutter && flutter test)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,6 @@ jobs:
         uses: actions/checkout@v3
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
-      - name: Install dependencies
-        run: dart pub get --directory packages/dart
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/dart
   check-code-analysis-flutter:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,14 +11,27 @@ on:
...
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup flutter
+        uses: subosito/flutter-action@v2
+        with:
+          cache: true
+      - name: Lint
+        run: flutter format --output=none --set-exit-if-changed packages/flutter
```
fischerscode, line 26: I think we have to check the format for the dart package, too. This should only check packages/flutter.
I see 3 options:
1. Copy the job (check-lint-dart with dart-lang/setup-dart@v1)
2. use `flutter format --output=none --set-exit-if-changed packages` to format both packages
3. Install flutter and dart in one job and run both format commands

.github/workflows/ci.yml
```diff
@@ -48,35 +48,17 @@ jobs:
...
-        run: (cd packages/dart && dart run build_runner build --delete-conflicting-outputs)
-      - name: Run tests
-        run: (cd packages/dart && dart test)
   check-flutter:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         channel: [stable]
+    name: Test flutter (${{ matrix.os }}, ${{ matrix.channel }})
```
fischerscode, line 57: Great, didn't know that's possible.

.github/workflows/ci.yml
```diff
@@ -11,52 +11,84 @@ on:
...
-      - name: Analyze project source
-        run: flutter analyze packages/flutter
       - name: Run tests
         run: (cd packages/flutter && flutter test)
+  check-dart:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        channel: [stable]
```
fischerscode, line 92: We could set strategy.fail-fast to false in order to keep the other tests running if one of the matrix fails.

mtrezza: added

.github/workflows/ci.yml
mtrezza: Yes, makes it a bit more readable

.github/workflows/ci.yml
mtrezza: I've added a separate job, somehow feels more "correct" to let dart to its own analysis.

mtrezza:
- I'm not happy with the lint checks, they don't output what exactly they have found. Makes it hard to fix these.
- What does the code analysis need? I thought it's only checking the source code, or does this need to run after compile?

fischerscode:
```
Changed packages/flutter/example/test/data/repository/diet_plan/repository_diet_plan_db_test.dart
Changed packages/flutter/example/test/data/repository/diet_plan/repository_diet_plan_test.dart
Changed packages/flutter/example_livelist/lib/main.dart
Formatted 34 files (3 changed) in 0.84 seconds.
Formatting failed: 1
Error: Process completed with exit code 1.
```
They tell which files changed, I think it's more a "you missed auto formatting the code".
Alternatively it might be possible to let actions reformat the code and push the changes to the PR.

fischerscode:
> What does the code analysis need?

I think it needs all dependencies being resolved, since it also checks if the types match.
For the flutter package this means running pub get in the dart package is required.

mtrezza:
I see; for npm lint it is not always able to auto-fix everything, so it's essential to see the exact error. But if this can auto-fix everything then I guess we are good here.

fischerscode:
@mtrezza, pub get is also needed in the package where the tests are run. Just tested it.

fischerscode:
> If we add a CONTRIBUTING.md file here with FAQ to explain what to do, then that'd also do it I think.

👍

mtrezza:
Should I replace

```
(cd packages/dart && flutter pub get)
cd packages/flutter
```

with
```
(cd packages/flutter &&flutter pub get)
(cd packages/dart &&dart pub get)
```

mtrezza:
> pub get is also needed in the package where the tests are run. Just tested it.

Could you do a review and let me know what I need to change in the review tab?

.github/workflows/ci.yml
```diff
@@ -11,52 +11,104 @@ on:
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+      fail-fast: false
+    name: Test dart (${{ matrix.os }})
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup dart
+        uses: dart-lang/setup-dart@v1
+        with:
+          sdk: 'stable'
```
fischerscode, line 108: stable should be the default, but specifying it explicitly should be fine, too.

.github/workflows/ci.yml
```diff
@@ -11,52 +11,104 @@ on:
...
+    name: Test flutter (${{ matrix.os }})
     steps:
-      - name: Checkout the code
+      - name: Checkout code
         uses: actions/checkout@v3
       - name: Setup flutter
         uses: subosito/flutter-action@v2
         with:
-          channel: ${{ matrix.channel }}
+          channel: 'stable'
```
fischerscode, line 84: stable should be the default, but specifying it explicitly should be fine, too.

.github/workflows/ci.yml
```diff
@@ -11,52 +11,104 @@ on:
...
+    name: Code analysis (dart)
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Setup dart
+        uses: dart-lang/setup-dart@v1
+      - name: Analyze code
         run: dart analyze packages/dart
```
fischerscode, line 70: `dart pub get --directory packages/dart` must be run before analyze

.github/workflows/ci.yml
```diff
@@ -11,52 +11,104 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
         uses: actions/checkout@v3
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
-        with:
-          sdk: ${{ matrix.sdk }}
       - name: Install dependencies
         run: dart pub get --directory packages/dart
```
fischerscode, line 37: That should not be needed.

fischerscode (changes requested):
Almost ready 👍

.github/workflows/ci.yml
mtrezza: removed

.github/workflows/ci.yml
mtrezza: removed

.github/workflows/ci.yml
mtrezza: added

.github/workflows/ci.yml
mtrezza: removed

.github/workflows/ci.yml
```diff
@@ -11,52 +11,101 @@ on:
...
+      - name: Checkout code
         uses: actions/checkout@v3
       - name: Setup flutter
         uses: subosito/flutter-action@v2
         with:
-          channel: ${{ matrix.channel }}
+          cache: true
       - name: Install dependencies
         run: |
           (cd packages/dart && flutter pub get)
```
fischerscode, line 87: Just realized, that's not working on windows.
Windows does not change back in the previous directory like linux/macos does :(

Replacing this line with
```
cd packages/dart
flutter pub get
cd ../..
```
should work.

fischerscode (changes requested):
By looking through the ci logs, I've discovered a problem regarding special behavior on windows. Apart from that, it looks good to me.

.github/workflows/ci.yml
```diff
@@ -11,52 +11,105 @@ on:
...
+        uses: actions/checkout@v3
+      - name: Setup flutter
+        uses: subosito/flutter-action@v2
         with:
-          sdk: ${{ matrix.sdk }}
+          cache: true
+      - name: Install dependencies
+        run: |
+          cd packages/dart
+          flutter pub get
```
mtrezza, line 52: @fischerscode shouldn't this be `dart pub get`?

.github/workflows/ci.yml
mtrezza: changed

.github/workflows/ci.yml
fischerscode: It should work.
I think dart would work, too. But I'm not sure if subosito/flutter-action also installs the dart command.

mtrezza: OK so I leave it as it is?

fischerscode: I would leave it like this.

fischerscode (approved):
Looks good to me.

PR #724, opened 2022-04-30
refactor: update example app to use Android embedding v2 

fischerscode:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Issue Description
Recreated the android application of the flutter example to fix `pub get`.

Related issue: #722 

### Approach
Just recreated...

### TODOs before merging
- [ ] Maybe adjust the title, I had no idea how to name it.

### Expected behavior
The test should fail at the formatting step.

PR #731, opened 2022-04-30
style: change code analyzer configuration to pass in CI

fischerscode:

### Issue Description
This PR revisits the linter/analyzer configuration and hopefully makes the tests green.

Related issue: #729
Close: #729

### Approach
I decided to
- use "no-fatal-infos" (Fixing all infos is something for the future.)
- set "publish_to: none" when analyzing the flutter package while using the path dependency
- fix the sdk of example_livelist (this has likely been missed while updating to null-safety)
- use the recommended lint settings. (The project used the settings from the flutter project back when we split the package into the dart and flutter part. Since there are now recommended and easy to use settings, why not use them. Esp. since, as far as I know, the settings where never customized. (Keep it simple!)) (@mtrezza "DO use curly braces for all flow control structures." is a info now)

Feedback is more than welcome.

### TODOs before merging
n/a


fischerscode:
> ### Thanks for opening this pull request!
> 
>     * 🎉 We are excited about your hands-on contribution!

Didn't know just `Close:` is not enough.

fischerscode:
Expected behavior: some tests should fail, since #724 is not merged, yet.

.github/workflows/ci.yml
```diff
@@ -52,11 +52,12 @@ jobs:
           flutter pub get
           cd ../..
           cd packages/flutter
+          echo "publish_to: none" >> pubspec.yaml
```
mtrezza, line 55: What does this do?

packages/dart/pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dependencies:
   mime_type: ^1.0.0
 
 dev_dependencies:
+  lints: ^1.0.1
```
mtrezza, line 27: Should we remove the range operators for consistency?

mtrezza:
What is the diff between the code analyzer and lint stage in the CI? I see the config file in this PR also includes lint rules, which makes me think we may only need one of them but not both?

.github/workflows/ci.yml
fischerscode: flutter analyze failes if the dart package is added using path since it would not work when publishing the package. This trick tells the analyzer we don't want to publish the package. (What is actually not true.)

packages/dart/pubspec.yaml
fischerscode: In order to make sure the tests always run using the same versions?
We could do that, but for the dependencies I would keep them, since dependency hell would strike soon.

fischerscode:
> What is the diff between the code analyzer and lint stage in the CI? I see the config file in this PR also includes lint rules for the analyzer, which makes me think we may only need one of them but not both?

Analyzer makes sure the code is valid, linter enforces a coding style. But I think this is pretty technical, for the average developer (including me) it is the same working together in order to tell what's wrong.
https://stackoverflow.com/a/61931536/11456894

packages/dart/pubspec.yaml
mtrezza: It would be best to keep it fixed and use an external service to upgrade it on demand, like Snyk does for npm dependencies. Unfortunately it doesn't support dart dependencies. Do you know of any 3rd party service for dart?

mtrezza:
Got it, thanks, so let's keep both.

fischerscode:
Its green! 🎉

.github/workflows/ci.yml
mtrezza: Cool!

mtrezza:
Amazing! You did a great job!

packages/dart/pubspec.yaml
fischerscode: Do you want to remove the range operators just for the dev_dependencies or for all dependencies?
I would not remove them for all, since it gives `pub get` more options for finding versions that meet all requirements.

Regarding a tool, I don't know one, that supports it, but I already thought about adding such a functionality to [uptodate](https://github.com/fischerscode/uptodate).

fischerscode:
Well, I would consider it to be team work. 😃

packages/dart/pubspec.yaml
mtrezza: Could you explain how the lock file works here, is that updated on every publish or is it a distinct command that we should run from time to time to keep the dependencies updated at least within the range that is specified in `pubspec.yaml`?

mtrezza (approved):
Looks good!

packages/dart/pubspec.yaml
mtrezza: I guess this is a separate discussion unrelated to this PR, so I'll resolve this for now.

fischerscode: TLDR;
Pub checks what are the best versions it can use.

Pub is the package manager of dart/flutter. "pubspec.yaml" sets the the dependencies the project needs. The dependencies are the ones the code uses. dev_dependencies are used for tests and stuff during development.
Packages, like this project, usually don't version control the 'pubspec.lock' file. Applications usually do.
When `pub get` runs, it determines the latest versions for all the (nested) dependencies. Since the versions are usually specified like `package: ^1.2.3`, which means every stable version greater or equal to 1.2.3 but smaller then 2.0.0, pub has some room to update the dependencies without running into breaking changes. This is needed, since packages might share dependencies and not every package is 100% up to date.
The versions used get stored in `pubspec.lock`. The next time `pub get` runs, it just checks if `pubspec.lock` still meets the requirements of `pubspec.yaml`. If not, versions are determined again.
`pub upgrade` can be used to force updates to `pubspec.lock`.
`pub outdated` can list possible updates (including major changes).

Additionally the developer has the option to force override a dependency, even if it does not satisfy the requirements of other (nested) dependencies.

mtrezza: > The next time pub get runs, it just checks if pubspec.lock still meets the requirements of pubspec.yaml. If not, versions are determined again.

Let's assume this scenario:

- dependency version in pubspec.yaml: ^1.0.0
- dependency version in pubspec.lock: 1.0.1
- latest dependency version available: 1.0.2

Then I publish a release and the workflow runs `pub get`, what happens? Is pubspec.lock updated to 1.0.2 or does it stay 1.0.1?

PR #736, opened 2022-05-01
docs: update changelogs

mtrezza:

### Issue Description
- Changelogs are restyled according to Parse Org-wide style
- Changelog entry for release 3.1.1 is added

Related issue: #668 

### Approach
n/a

### TODOs before merging
n/a

### DIFF
--- a/packages/dart/CHANGELOG.md
+++ b/packages/dart/CHANGELOG.md
@@ -1,110 +1,151 @@
-## 3.1.0
-Bug fixes
-General improvements
-Updated dependencies
+# [3.1.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/V3.0.0...V3.1.0) (2021-06-28)
 
-## 3.0.0
-Stable null safety release.
+### Bug Fixes
 
-## 2.1.0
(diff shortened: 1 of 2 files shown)


mtrezza:
@fischerscode I don't understand the changelogs; the logs for dart and flutter are identical from 2.0.0 onwards, but these are 2 different packages. How come?

fischerscode:
@mtrezza Originally, there was only one package. Then it got split in the dart and the flutter package. In order to make the changes of the dart package [transparent](https://pub.dev/packages/parse_server_sdk_flutter/changelog) to the users of the flutter one, changes where added to both. The two packages where treated as one.
But it might be better to stop doing it like this.

mtrezza:
Yes, it seems weird. I can also imaging that there may be a change in the Flutter package that has nothing to do with the darf package, but it would appear in the dart changelog. I am beginning to doubt whether we should really keep the versions in sync.

mtrezza:
I've added version 3.1.1 only to the flutter changelog, since it's the flutter example app, unrelated to dart.

fischerscode (commented):
LGTM

fischerscode (approved):
LGTM (clicked the wrong button)

Issue #799, opened 2022-11-21
healthCheck throws error


catalunha:
# I'm building a tutorial for Parse Serve (using Back4app) and ran into this issue on SKD for Flutter.

### New Issue Checklist
- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [ x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
I tested the following code:
```dart
void main(List<String> arguments) async {
  // String appId = 'appIdCorrect';
  // String clientKey = 'clientKeyCorrect';
  String appId = 'appIdWrong';
  String clientKey = 'clientKeyWrong';
  String keyParseServerUrl = 'https://parseapi.back4app.com/';
  var parse = await Parse().initialize(
    appId,
    keyParseServerUrl,
    clientKey: clientKey,
    autoSendSessionId: true,
    debug: true,
  );
  print(
      'Parse().hasParseBeenInitialized(): ${parse.hasParseBeenInitialized()}');
  if (parse.hasParseBeenInitialized()) {
    var response = await parse.healthCheck();
    print('response.success: ${response.success}');
    print('Bach4App: Ok');
  } else {
    print('Bach4App: Not Ok');
  }
}
```

### Steps to reproduce
Create a pure standard Dart project with VSCode
Create a basic project in back4app and write down the appId and clientKey.
Copy the code above to your Dart project and enter the correct appId and clientKey and then a wrong value.
Run the Dart project with:
$ dart bin/learning_about_b4a_dart.dart 

### Actual Outcome

If I put the correct appId and clientKey values I get:
Parse().hasParseBeenInitialized(): true
response.success: true
Bach4App: Ok

But if I put the wrong appId and clientKey values I get:
Parse().hasParseBeenInitialized(): true
Unhandled exception:
type 'Null' is not a subtype of type 'int'
#0      buildErrorResponse (package:parse_server_sdk/src/objects/response/parse_error_response.dart:8:25)
[...]

Issue #487, opened 2020-10-14
signUp() returns null or "bad or missing username"


ernestasga:
When I try to sign up with my User class which extends ParseUser
```
    var res = await user.signUp();
    print(res);
```
 it returns null, but if I call it from new instance

```
    user = User('username', 'password', 'email@fake.com');
    var res = await user.signUp();
    print(res);
```
it end with bad or missing username

```
I/flutter ( 4763): ╭-- Parse Request
I/flutter ( 4763): curl -X POST -H 'X-Parse-Revocable-Session: 1' -H 'X-Parse-Installation-Id: ffccd1b3-6ea4-4d9f-a3d0-fe0f4d5eecc3' -H 'content-type: null' -H 'user-agent: Flutter Parse SDK 2.0.0' -H 'X-Parse-Application-Id: com.ernestasg.khons' http://10.0.2.2:1337/parse/classes/_User
I/flutter ( 4763):
I/flutter ( 4763):  http://10.0.2.2:1337/parse/classes/_User
I/flutter ( 4763): ╰--
I/flutter ( 4763): ╭-- Parse Response
I/flutter ( 4763): Class: _User
I/flutter ( 4763): Function: ParseApiRQ.signUp
I/flutter ( 4763): Status Code: -1
I/flutter ( 4763): Type: UnknownError
I/flutter ( 4763): Exception: DioError [DioErrorType.RESPONSE]: Http status error [400]
I/flutter ( 4763): ╰--
I/flutter ( 4763):
I/flutter ( 4763): Instance of 'ParseResponse'
```

Also I'm using getit for dependency injection. When using `registerLazySingleton()`  I'm unable to logout after logging in second time because sessionToken is null. My workaround was to use `registerFactory()` and get new instance when logging in `this.user = sl.get<User>();`, maybe there is a better way to handle this?

Issue #481, opened 2020-10-13
Problems to recover error messages after switching to the Dio package.


RodrigoSMarques:
I identified two problems in version 1.0.28 that do not happen in 1.0.27.

The problem is related to switching to the Dio package.

1) When executing a function, I use try / catch to retrieve the error message. The error message (String) is no longer being returned. A DioError is being returned.

Version 1.0.27 - expected response
```dart
flutter: ╭-- Parse Request
flutter: curl -X POST -H 'content-type: text/plain; charset=utf-8' -H 'user-agent: Flutter Parse SDK 1.0.27' -H 'X-Parse-Application-Id: xxxxxxxxxxx' -H 'X-Parse-Session-Token: r:xxxxxxxxxxx' -H 'X-Parse-Client-Key: xxxxxxxxxxx' -d '{"code":"TESTE10","minOrderValue":100.0,"discountValue":10.0,"qtyTotal":9999,"expirationDate":1602817140000,"advertiserID":"iUAzCRbUcQ"}' https://xxx.xxxxxxx.com/1/functions/saveCoupon

 https://xxx.xxxxxxx.com/1/functions/saveCoupon
flutter: ╰--
flutter: ╭-- Parse Response
Class: saveCoupon
Function: ParseApiRQ.execute
Status Code: 141
Type: ScriptError
Error: Coupon Code in use. Try another code
╰--
```
Version 1.0.28 - error
```dart
flutter: ╭-- Parse Request
flutter: curl -X POST -H 'content-type: text/plain; charset=utf-8' -H 'user-agent: Flutter Parse SDK 1.0.28' -H 'X-Parse-Application-Id: xxxxxxxxxx' -H 'X-Parse-Session-Token: r:xxxxxxxxxxx' -H 'X-Parse-Client-Key: xxxxxxxxxxx' -d '{"code":"TESTE10","minOrderValue":100.0,"discountValue":10.0,"qtyTotal":9999,"expirationDate":1602817140000,"advertiserID":"iUAzCRbUcQ"}' https://xxx.xxxxxxx.com/1/functions/saveCoupon
 https://xxx.xxxxxxx.com/1/functions/saveCoupon
flutter: ╰--
[VERBOSE-2:ui_dart_state.cc(177)] Unhandled Exception: type 'DioError' is not a subtype of type 'String'
```

2) When an error occurs when saving an object and an exception is generated, the error is not being returned.

Version 1.0.27 - expected response
```dart
[...]

PR #802, opened 2022-11-27
fix: SDK crashes due to missing error code property in `ParseNetworkResponse.data`

mbfakourii:

### Issue Description
sometimes in buildErrorResponse ```code``` not exist in ```apiResponse.data``` and ```json.decode``` can't parse it!
but statusCode exist in ParseNetworkResponse and can be taken from this variable.

Related issue: https://github.com/parse-community/Parse-SDK-Flutter/issues/799
Closes: https://github.com/parse-community/Parse-SDK-Flutter/issues/799

### Approach
In some times when the ```code``` is not available in ```ParseNetworkResponse.data```, the ```statusCode``` is used in the ```ParseNetworkResponse.statusCode``` variable.

### TODOs before merging
- [ ] Add tests
- [ ] A changelog entry


mtrezza:
Which side effects could that change have?

mbfakourii:
> What side effects could that change have? The way of building the error response with `responseData[keyCode]` has worked so far it seems?

I did various tests and found out that sometimes the ```code``` is not sent. That's why I put the null checker ```??``` for when it is not available, and when the ```code``` is not sent, it uses ```ParseNetworkResponse.statusCode```.

mtrezza:
So the underlying issue is that Parse Server sometimes doesn't send that code?

mbfakourii:
> So the underlying issue is that Parse Server sometimes doesn't send that code?

yes for example 
if ```appID``` is wrong and try for get ```health```
```https://parseapi.back4app.com/health```
Json received without ```code``` is like this
```json
{"error":"unauthorized"}
```

mtrezza:
Got it, I'm a bit careful about introducing this ambiguity. It basically means that a developer cannot be sure whether the code is a parse error code or a http response code, right?

For example, if the error code is 400 and Parse Server also has a Parse Error code 400, how should a developer interpret this, and how should error handling logic in an app deal with this?

mbfakourii:
> Got it, I'm a bit careful about introducing this ambiguity. It basically means that a developer cannot be sure whether the code is a parse error code or a http response code, right?
> 
> For example, if the error code is 400 and Parse Server also has a Parse Error code 400, how should a developer interpret this, and how should error handling logic in an app deal with this?
> 
> Maybe an alternative would be to use a distinct generic Parse Error code if no specific error code is present.
> 
> For example the [docs](https://docs.parseplatform.org/js/guide/#other-issues) already offer `OtherCause` (-1): "An unknown error or an error unrelated to Parse occurred."

It is true that the developer is not able to understand the meaning of the code.
I use ```-1``` instead of ```ParseNetworkResponse.statusCode```

packages/dart/lib/src/objects/response/parse_error_response.dart
```diff
@@ -5,7 +5,8 @@ ParseResponse buildErrorResponse(
     ParseResponse response, ParseNetworkResponse apiResponse) {
   final Map<String, dynamic> responseData = json.decode(apiResponse.data);
   response.error = ParseError(
-      code: responseData[keyCode], message: responseData[keyError].toString());
-  response.statusCode = responseData[keyCode];
+      code: responseData[keyCode] ?? -1,
+      message: responseData[keyError].toString());
+  response.statusCode = responseData[keyCode] ?? -1;
```
mtrezza, line 10: Do we have a list of Parse Errors in this SDK anywhere? These error codes are usually defined somewhere. See for example [ParseError](https://github.com/parse-community/Parse-SDK-JS/blob/alpha/src/ParseError.js) in the Parse JS SDK.

If the Flutter SDK doesn't have that, could you please centrally defined at least the `-1` error code with the same name as in the Parse JS SDK, and use that reference instead of directly returning `-1`.

mtrezza (changes requested):

packages/dart/lib/src/objects/response/parse_error_response.dart
mbfakourii: In the Flutter SDK, we have a list of errors in this [file](https://github.com/parse-community/Parse-SDK-Flutter/blob/d5f0657dd06fbb4ad6c7890303e946d72d39e773/packages/dart/lib/src/objects/parse_error.dart#L17), but the variable is private and cannot be used in other classes. I will try to improve this section.

mbfakourii:
Changed and added const error code based on [Parse-SDK-JS](https://github.com/parse-community/Parse-SDK-JS/blob/5c140299109d2850982155ac1e1b907e1895fde0/src/ParseError.js)

packages/dart/lib/src/objects/response/parse_error_response.dart
mtrezza: I think it should be public anyway, because a developer may want to reference the error code somewhere in their code to identify the error.

mbfakourii: > Amazing! Could you please preserve the numeric codes as well? It seems that in the commit they were removed.

all numeric codes are linked to const keys. like [this](https://github.com/parse-community/Parse-SDK-Flutter/blob/60a82288aa329b0537b5d936e6c5785837294be7/packages/dart/lib/src/objects/parse_error.dart#L228).
what codes do you mean ?

mtrezza: Looking at it now it looks fine, I think I only saw a truncated version when looking at the changes previously.

packages/dart/lib/src/objects/parse_error.dart
```diff
@@ -13,58 +13,279 @@ class ParseError {
...
+  /// we had to use an XDomainRequest object to allow CORS requests in
+  /// Internet Explorer, which strips the body from HTTP responses that have
+  /// a non-2XX status code.
+  static const int xDomainRequest = 602;
+
   static const Map<int, String> _exceptions = {
-    -1: 'UnknownError',
-
-    // SDK errors / Errors
-    1: 'No Results',
```
mtrezza, line 20: It seems that these error codes (which you already removed) were actually some Flutter SDK-custom codes (1, 2) and a HTTP response code (400).

Have these 3 codes been used anywhere? If so, this is a breaking change, at least for code 1 which was described as "No Results" but is now "internalServerError".

mtrezza: I don't think it makes sense to return them, the 400 code is incorrect anyway, because it mixes a HTTP response code into the Parse Error codes. If you are 100% sure they weren't used anywhere we can just remove them. Did you also do a simple text search to be sure?

mbfakourii: yes 
In addition, I saw error 400 in several issues, none of which were related to code 400 https://github.com/parse-community/Parse-SDK-Flutter/issues/487#issue-721516490  ,https://github.com/parse-community/Parse-SDK-Flutter/issues/487#issue-721516490 and https://github.com/parse-community/Parse-SDK-Flutter/issues/481#issue-719863736

contributors parse_error.dart are codes 1, 2 and 400 important?
@nstrelow 
@fischerscode

mbfakourii: @mtrezza 
I suggest not to add codes unless there is a problem because I don't see any use of these codes in Parse Flutter!

mtrezza (dismissed):
Looks good to me; unless we get contrary feedback regrading the 3 removed codes.

- Is the PR title correct?
- Could you add the changelog entry by copy/pasting the PR title + bump the version for the release?

mbfakourii:
> @mbfakourii You have been quite active recently and make some great contributions. Can I add you to the review team of the Parse Flutter SDK? You would get notified if there are PRs to review.

Yes, I will be happy to add

 and title added in changelog

packages/flutter/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: parse_server_sdk_flutter
 description: Flutter plugin for Parse Server, (https://parseplatform.org), (https://back4app.com)
-version: 3.1.3
+version: 3.1.4
```
mtrezza, line 3: Why is the flutter version being bumped if the change is only in the dart package?

See https://github.com/parse-community/Parse-SDK-Flutter/commit/7fb7ce67d5bea30d583f4b182303ee6a31214185 where we also only had a change in the dart package and only the dart version was bumped.

packages/dart/lib/src/base/parse_constants.dart
```diff
@@ -1,7 +1,7 @@
 part of flutter_parse_sdk;
 
 // Library
-const String keySdkVersion = '3.1.2';
+const String keySdkVersion = '3.1.4';
```
mtrezza, line 4: Did you manually change this version or is this auto-generated?

mbfakourii: > Did you manually change this version or is this auto-generated?

manually

packages/flutter/pubspec.yaml
mbfakourii: It was my mistake, I thought it was for dart

mbfakourii: Returned to the previous version

packages/dart/lib/src/base/parse_constants.dart
mtrezza: We haven't changed that in the last release, maybe by mistake?

I think we should create a list where the version has to be changed for a new release.

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [3.1.4](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.3...dart-3.1.4) (2022-12-10)
+
+### Bug Fixes
+
+* fix: SDK crashes due to missing error code property in `ParseNetworkResponse.data` ([#799](https://github.com/parse-community/Parse-SDK-Flutter/issues/799))
```
mtrezza, line 5:
```suggestion
* SDK crashes due to missing error code property in `ParseNetworkResponse.data` ([#799](https://github.com/parse-community/Parse-SDK-Flutter/issues/799))
```

packages/flutter/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
   flutter:
     sdk: flutter
 
-  parse_server_sdk: ^3.1.2
+  parse_server_sdk: ^3.1.3
```
mtrezza, line 13: I think this also needs to be reverted, otherwise we'd have to release a new flutter version?

packages/dart/lib/src/base/parse_constants.dart
mbfakourii: > We haven't changed that in the last release, maybe by mistake?

Yes, in the previous version, I didn't know exactly that I had to change the parse_constants file

> I think we should create a list where the version has to be changed for a new release. So for dart it's in:
> * packages/dart/lib/src/base/parse_constants.dart
> * packages/dart/pubspec.yaml
> 
> ?

Yes, it's a good idea

packages/flutter/pubspec.yaml
mbfakourii: No, there is no need Returned to the previous version

packages/dart/lib/src/base/parse_constants.dart
mtrezza: Why is this variable here needed at all?

mbfakourii: This variable is sent to the server as the version of the requester in a header in keys ```parseVersion``` and ```user-agent```

and It is not a critical variable, but it is good to change it

mtrezza: So this was actually bug in the previous version then. Is there any way we could get the package version from the package itself instead of manually maintaining this var?

mbfakourii: > So this was actually bug in the previous version then. 

yes
> Is there any way we could get the package version from the package itself instead of manually maintaining this var?

Yes, but you have to use other packages [package_info_plus](https://pub.dev/packages/package_info_plus) and this package does not support the Dart version, it only supports Flutter!

mbfakourii: @mtrezza 
Is it possible to Merge this PR? I need these codes in another PRs.

mtrezza:
Off-topic; if you're interested please feel free to join our Slack workspace. There are actually a few Flutter related developer questions that came in today, and you seem to be a Flutter expert, so this could be a good match.

packages/dart/lib/src/base/parse_constants.dart
mtrezza: Sure, let' keep this in mind, maybe we can find a way to replace that extra var in the future...

packages/dart/lib/src/objects/parse_error.dart
mtrezza: Sounds good

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [3.1.4](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.3...dart-3.1.4) (2022-12-10)
```
```suggestion
## [3.1.4](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.3...dart-3.1.4) (2022-12-14)
```

mtrezza (approved):
Looks good!

Issue #821, opened 2023-02-12
Upgrade dependency package_info_plus


sgehrman:
### New Issue Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [ x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description

parse_server_sdk_flutter >=3.1.3 depends on `package_info_plus: ^1.4.2`.
The dependency should be upgraded as version `3.0.3` is already available.

mtrezza:
Could you give more details about the issue?

sgehrman:
weird response, I'm just asking for you to update packages.  Is this difficult to understand?

mtrezza:
Thanks, I will rephrase the issue for you, so you can see the difference. I will clean your post from comments like "lame", "stupid", etc. - could you clarify what you mean by that? Is there another issue what you would like to express in addition to the update?

tungcyfeer:
Hi. I have the same issue, because I use `package_info_plus` with version 3.x in my own package. I just checked your lib, it has ^1.4.2 in yaml, so I got version solving failed. So I think you can consider upgrade `package_info_plus` to the latest version when upgrade version of this lib. Thanks!

sgehrman:
still a problem in October.  This is the worst maintained package on Flutter!!

PR #824, opened 2023-02-23
feat: Upgrade various dependencies and fix warnings

mbfakourii:

### Issue Description
All versions and packages have been updated to the latest version and warnings have been resolved

Closes: #821

### Approach
n/a

### TODOs before merging
- [ ] Add tests
- [ ] A changelog entry


mbfakourii:
@mtrezza 
I updated and tested the packages, but CI is trying to use old packages!

In the Flutter package in this [line](https://github.com/parse-community/Parse-SDK-Flutter/blob/b07ce8c7fbe709333a6a97ce5473e6d5c88f33f0/tools/flutter-dependencies.sh#L10), it uses the internal Dart package, which runs into problems because it uses new packages.

mtrezza:
There are quite some changes in this PR, are they all related to the info package?

mtrezza:
Why is it trying to use old pkgs?

mbfakourii:
> Why is the CI trying to use old pkgs?

For example, the Dart package uses dio: ^5.0.0, but the Flutter package uses dio: ^4.0.6, and we also use the parse_server_sdk: ^3.1.13 package, and this version uses dio: ^4.0.6. They have conflict. In CI, which is in this section, it uses parse_server_sdk at the local level and it is completely in the configuration.

mbfakourii:
My suggestion is to first give a new version in the Dart version with dio: ^5.0.0 and then give this new version for the Flutter version. Regardless of CI errors

mtrezza:
This is how we'd usually do it, as you described:

1. Release dart with upgrade to dio 5
2. Release flutter with upgrade to dio 5 and upgrade to the new dart release with dio 5

That should work fine. The question is why do we need the script that uses the local version of dart?
I imagine the script is for developing and testing changes locally that span across dart and flutter. But on GitHub in the CI I think that script should never run because a PR and its CI should only look at an individual release which can either be a dart or a flutter release, never both.

So the solution could be to always avoid running that script in the CI. Where is that script called?

mbfakourii:
> This is how we'd usually do it, as you described:
> 
> 1. Release dart with upgrade to dio 5
> 2. Release flutter with upgrade to dio 5 and upgrade to the new dart release with dio 5
> 
> That should work fine. The question is why do we need the script that uses the local version of dart? I imagine the script is for developing and testing changes locally that span across dart and flutter. But on GitHub in the CI I think that script should never run because a PR and its CI should only look at an individual release which can either be a dart or a flutter release, never both.
> 
> For example, if a new feature requires changes in dart and flutter, both changes can be developed locally. The developer wants to run tests locally using the script. Then a PR needs to be opened for dart, and a separate PR for flutter. In both PRs the script should not run in the GitHub CI.
> 
> So the solution could be to never run the script in the CI but provide it for the developer to manually run locally. Where is that script called?

I agree with you, there should not be two ways, In my opinion too, this should be left to the developer 

[In this script](https://github.com/parse-community/Parse-SDK-Flutter/blob/b07ce8c7fbe709333a6a97ce5473e6d5c88f33f0/tools/flutter-dependencies.sh), in line 9, the library corresponding to the pubdev version is deleted and in line 10, the local library is connected, which will check the local version

mbfakourii:
I bumped the Dart version, I think it is ready for merge
After merge, I will increase the flutter version in another PR

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [3.1.14](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.13...dart-3.1.14) (2023-02-25)
```
mtrezza, line 1: If we upgrade to dio 5, isn't this a breaking change for this SDK? The Parse Flutter SDK references this `^3.1.2` then there will be a conflict. So we need to release the dart package as `4.0.0`, right?

packages/flutter/pubspec.yaml
```diff
@@ -10,29 +10,29 @@ dependencies:
   flutter:
     sdk: flutter
 
-  parse_server_sdk: ^3.1.2
```
mtrezza, line 13: There can be no changes in the flutter package in a dart package release.

mtrezza (changes requested):

packages/dart/CHANGELOG.md
mbfakourii: No, I don't think it is a breaking change because we are only updating the packages and there are no changes at the api level.
I don't think package conflict is a breaking change. Because you can still use the old package.

packages/flutter/pubspec.yaml
mbfakourii: This change does not affect the Flutter codes, and after the release of the Dart version, the Flutter version will be released immediately

mtrezza: Still this is a dart release, so there cannot be any changes in the flutter package.

packages/dart/CHANGELOG.md
mtrezza: Let's assume an app has the dependency `flutter: ^3.1.3`. That flutter version has the dependency `parse_server_sdk: ^3.1.2`. What happens on the next `flutter pub get`? If I understand it correctly, it will pull `flutter: 3.1.3` with dio 4 and that will pull the latest non-breaking release of `parse_server_sdk: 3.1.14` with dio 5. Is that correct?

mbfakourii: Yes, it happens in the same way and there is no conflict

packages/flutter/pubspec.yaml
mbfakourii: No, it is not necessary. It can be added in another PR related to Flutter. I will return this part

packages/dart/CHANGELOG.md
mtrezza: Shouldn't there be a conflict if the flutter SDK uses dio 4 but its dependency dart SDK uses dio 5?

packages/flutter/pubspec.yaml
mbfakourii: I returned to the previous version

packages/dart/CHANGELOG.md
mbfakourii: > Shouldn't there be a conflict if the flutter package uses dio 4 but its dependency the dart package uses dio 5?

Yes, in this case we have a conflict

mtrezza: I don't see a way around this conflict. If we release the dart package as `3.x` then there's the conflict, so I guess releasing as 4.x is the only way around this, right?

mbfakourii: Yes, it is true, I think it would be better to upgrade to version ```4.0.0```

mtrezza: Alright, thanks for confirming

mtrezza:
Where do the errors in the CI come from? Is that because of the script that installs a local version of dart in the flutter package?

mtrezza (commented):
The coverage test is failing here. Could that be related to the dependency updates?

mtrezza:
Codecov is failing due to lower coverage it seems. We can still merge I guess.

- Bumped version to 4.0.0 due to breaking change with dio 5.x upgrade
- Added breaking change note to changelog

Is this ready for merge?

packages/flutter/pubspec.yaml
```diff
@@ -35,4 +35,4 @@ dev_dependencies:
   # Testing
   flutter_test:
     sdk: flutter
-  mockito: ^5.2.0
+  mockito: ^5.2.0
```
mtrezza, line 38:
```suggestion
  mockito: ^5.2.0

```

mtrezza (changes requested):
Codecov is failing due to lower coverage it seems. We can still merge I guess.

- Bumped version to 4.0.0 due to breaking change with dio 5.x upgrade
- Added breaking change note to changelog

There are some flutter changes left in:

- packages/flutter/lib/src/utils/parse_live_list.dart
- packages/flutter/lib/src/utils/parse_live_grid.dart
- packages/flutter/example_livelist/lib/main.dart

Could you revert these, so we can merge and release the dart package?

packages/dart/lib/src/network/dio_adapter_js.dart
```diff
@@ -1,9 +1,7 @@
-import 'dart:io';
-
-import 'package:dio/adapter_browser.dart';
+import 'package:dio/browser.dart';
 import 'package:dio/dio.dart';
 
-HttpClientAdapter createHttpClientAdapter(SecurityContext? securityContext) {
+HttpClientAdapter createHttpClientAdapter(dynamic securityContext) {
```
Nidal-Bakir, line 4: If we will not use the `securityContext` do you think it is better to make it `_`?
like this:
```dart 
HttpClientAdapter createHttpClientAdapter(_) { /* code*/}

```

packages/dart/lib/src/network/parse_dio_client.dart
```diff
@@ -1,13 +1,11 @@
...
 import 'package:dio/dio.dart' as dio;
 import 'package:parse_server_sdk/parse_server_sdk.dart';
 
 import 'dio_adapter_io.dart' if (dart.library.js) 'dio_adapter_js.dart';
 
 class ParseDioClient extends ParseClient {
-  ParseDioClient(
-      {bool sendSessionId = false, SecurityContext? securityContext}) {
+  // securityContext is SecurityContext
+  ParseDioClient({bool sendSessionId = false, dynamic securityContext}) {
```
Nidal-Bakir, line 8: Why the securityContext is dynamic?

what was wrong with `SecurityContext?` type

packages/dart/lib/src/network/parse_dio_client.dart
```diff
@@ -178,8 +178,7 @@ class ParseDioClient extends ParseClient {
 
 /// Creates a custom version of HTTP Client that has Parse Data Preset
 class _ParseDioClient with dio.DioMixin implements dio.Dio {
-  _ParseDioClient(
-      {bool sendSessionId = false, SecurityContext? securityContext})
+  _ParseDioClient({bool sendSessionId = false, dynamic securityContext})
```
Nidal-Bakir, line 181: Also here the securityContext is dynamic

packages/dart/lib/src/network/dio_adapter_js.dart
mbfakourii: In some parts, we need it variable

packages/dart/lib/src/network/parse_dio_client.dart
mbfakourii: I used dynamic to reduce memory in the library. It doesn't matter, if you think it should be returned, I will return it

Nidal-Bakir: What about the conditional import:

```dart 
import 'dio_adapter_io.dart' if (dart.library.js) 'dio_adapter_js.dart';
```
When the library run on the web the `dio_adapter_js.dart` will be used. 

see this [answer](https://stackoverflow.com/a/75155651)

packages/dart/lib/src/network/dio_adapter_js.dart
Nidal-Bakir: I suggest the file be like this:
 
```dart
import 'package:dio/adapter_browser.dart';
import 'package:dio/dio.dart';

HttpClientAdapter createHttpClientAdapter(_) {
  return BrowserHttpClientAdapter();
}
 
```
What do you think?

packages/dart/lib/src/network/parse_dio_client.dart
mbfakourii: I am currently using universal_io, do you think we should replace the above method?

packages/dart/lib/src/network/dio_adapter_js.dart
mbfakourii: Your suggestion is currently being used in this [file](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/dart/lib/src/network/dio_adapter_js.dart)

packages/dart/lib/src/network/parse_dio_client.dart
Nidal-Bakir: Even `universal_io` is still using `dart_io` in its imports for the web. take a look at [http_client.dart](https://github.com/dint-dev/universal_io/blob/3f23477b0976754a7c4e4976401316ae4055c585/lib/src/http_client.dart#LL45C13-L45C13) which in turn been exported in[ _exports_in_browser.dart](https://github.com/dint-dev/universal_io/blob/3f23477b0976754a7c4e4976401316ae4055c585/lib/src/_exports_in_browser.dart#LL18C3-L18C3)

the [exports](https://github.com/dint-dev/universal_io/blob/3f23477b0976754a7c4e4976401316ae4055c585/lib/io.dart#L24-L26)
```dart 
export 'src/_exports_in_vm.dart'
    if (dart.library.html) 'src/_exports_in_browser.dart'
    if (dart.library.js) 'src/_exports_in_nodejs.dart';
```
  
So keep it `dynamic` for now. until we find a proper way to get around this. because currently, the SDK is full of `dynamic` and `Object` types. that is not good for the long term.

Thanks for your work.

mbfakourii:
@mtrezza 
Do you think it is ready for merge?

mtrezza:
I think we can, I was a bit hesitant because we have 2 packages that we - until now - always have released when the commit was made on that specific package. Now we're making dart commits while flutter hasn't been released yet. It's a change of process but I think it should be fine.

Nidal-Bakir:
> I was a bit hesitant because we have 2 packages that we - until now - always have released when the commit was made on that specific package

why not when a tag is created?

I have a package that gets published automatically when I create a tag for a commit.

My [publish.yml](https://github.com/Nidal-Bakir/exponential_back_off/blob/main/.github/workflows/publish.yml)
``` yaml
name: Publish to pub.dev

on:
  push:
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+*"

jobs:
  publish:
    runs-on: ubuntu-latest

    permissions:
      id-token: write
      contents: write

    steps:
      - name: 📚 Git checkout
        uses: actions/checkout@v3

      - name: 🎯 Setup Dart
        uses: dart-lang/setup-dart@v1

      - name: 📦 Install dependencies
        run: dart pub get

      - name: 🚀 publish
        run: dart pub publish --force

      - name: 🚀 Create release from tag for github releases
        uses: ncipollo/release-action@v1.12.0
        with:
          tag: ${{ github.ref_name }}
```

Nidal-Bakir:
We can modify my [publish.yml](https://github.com/Nidal-Bakir/exponential_back_off/blob/main/.github/workflows/publish.yml) to fit this project

mbfakourii:
> I think we can, I was a bit hesitant because we have 2 packages that we - until now - always have released when the commit was made on that specific package. Now we're making dart commits while flutter hasn't been released yet. It's a change of process, so I was worried about the version tags, etc. But I think it should be fine, do you see any concerns?

I don't think so, because the Flutter version will be released after the Dart version, so there is no need to worry

mtrezza:
> I have a package that gets published automatically when I create a tag for a commit.

We have that as well, see our [workflow](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/.github/workflows/release-automated.yml).

In the long run we want to automate releasing like we already do with other repos (e.g. Parse Server). Since this here is a monorepo it's more complicated. Looking forward to monorepo release automation, I had some concerns, but let's release this for now, I think we should be good.

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## [4.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.14...dart-4.0.0) (2023-02-27)
```
```suggestion
## [4.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-3.1.14...dart-4.0.0) (2023-03-07)
```

mtrezza (approved):
Looks good!

Issue #791, opened 2022-09-29
Change parse local database (sembast) path in IOS to LibraryDirectory rather then ApplicationDocumentsDirectory.


Nidal-Bakir:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Current Limitation
The database (sembast) file "parse.db" can be viewed by the regular user and can be deleted.
that's not good in some cases where the app will store some images, videos, and documents in the app directory, so the user will open the app directory to browse these files and see the "parse.db" database file!!
Or in some cases where the user suddenly decided to delete the app directory the (sembast) "parse.db" file will be deleted and the user will open the app to be surprised that his login and any saved data in the database were deleted.

### Feature / Enhancement Description
I suggest using LibraryDirectory as a path for the database file in IOS.
Rather than the ApplicationDocumentsDirectory path.

We can do something like this:
``` dart
class CoreStoreDirectory {
  Future<String> getDatabaseDirectory() async {
     if (Platform.isIOS) {
        return (await getLibraryDirectory()).path;
    } else {
        return (await getApplicationDocumentsDirectory()).path;
    }
  }
```

here:
https://github.com/parse-community/Parse-SDK-Flutter/blob/14be107e602f30db8f278aceeba3a8ae9b022e86/packages/flutter/lib/src/storage/core_store_directory_io.dart#L3-L6

### Example Use Case

![Selection_091](https://user-images.githubusercontent.com/29513372/193069731-670836a7-99a0-4d9c-bf9d-313770f2ba9e.png)

mtrezza:
Moving the DB file path makes this is a breaking change. Can we add a simple migration algorithm that looks for the existing file at the old path and move it to the new path if it exists?

Nidal-Bakir:
@mtrezza yes, I am well aware of this! I point it out in my PR description [here](https://github.com/parse-community/Parse-SDK-Flutter/pull/826#issue-1599228874)

### First solution:
This solution will return the old path (ApplicationDocumentsDirectory) if a database file exists in it. otherwise, it will return the new path (LibraryDirectory).

[...]

PR #826, opened 2023-02-24
fix: Parse SDK internal database file `parse.db` is accessible for app user on iOS and may be accidentally deleted

Nidal-Bakir:

### Issue Description

Closes: [#791](https://github.com/parse-community/Parse-SDK-Flutter/issues/791)

### Approach
Using LibraryDirectory as a path for the database.

### TODOs before merging
 This is a breaking change for **_ios_** platform because the SDK now will look for the "parse.db" in the `LibraryDirectory` rather than the `ApplicationDocumentsDirectory`.

We can support the old directory path by doing something like:
``` dart
    if (Platform.isIOS) {
      if(await _isDatabaseFileExistsInOldPath()){
        return (await path_provider.getApplicationDocumentsDirectory()).path;
      }
      return (await path_provider.getLibraryDirectory()).path;
    } else {
      return (await path_provider.getApplicationDocumentsDirectory()).path;
    }
  }

  Future<bool> _isDatabaseFileExistsInOldPath() async {
    final oldIosDatabaseDirPath =
        (await path_provider.getApplicationDocumentsDirectory()).path;
    final oldDatabaseFile =
        path.join('$oldIosDatabaseDirPath/parse', 'parse.db');

    if (await File(oldDatabaseFile).exists()) {
      return true;
    }
    return false;
  }
```

- [x] Add tests
- [x] A changelog entry

### DIFF
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -15,12 +15,9 @@ class CoreStoreDirectory {
     return (await path_provider.getApplicationDocumentsDirectory()).path;
   }
 
-  /// Why do we need this migration on iOS? see:
+  /// Migrate SDK internal database file on iOS, see:
   /// https://github.com/parse-community/Parse-SDK-Flutter/issues/791
-  ///
-  /// TODO: remove this migration algorithm after at least two major releases
-  ///
-  /// the release when this migration added is `3.1.4`
+  /// TODO: Remove this migration algorithm in the future. 
   Future<void> _migrateDBFileToLibraryDirectory() async {
     final dbFile = await _getDBFileIfExistsInAppDocDir();
 
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -17,7 +17,7 @@ class CoreStoreDirectory {
 
   /// Migrate SDK internal database file on iOS, see:
   /// https://github.com/parse-community/Parse-SDK-Flutter/issues/791
-  /// TODO: Remove this migration algorithm in the future. 
+  /// TODO: Remove this migration algorithm in the future.
   Future<void> _migrateDBFileToLibraryDirectory() async {
     final dbFile = await _getDBFileIfExistsInAppDocDir();
 
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -1,11 +1,17 @@
-import 'package:path_provider/path_provider.dart';
+import 'dart:io';
+
+import 'package:path_provider/path_provider.dart' as path_provider;
 
 class CoreStoreDirectory {
   Future<String> getDatabaseDirectory() async {
-    return (await getApplicationDocumentsDirectory()).path;
+    if (Platform.isIOS) {
+      return (await path_provider.getLibraryDirectory()).path;
+    } else {
+      return (await path_provider.getApplicationDocumentsDirectory()).path;
+    }
   }
 
   Future<String?> getTempDirectory() async {
-    return (await getTemporaryDirectory()).path;
+    return (await path_provider.getTemporaryDirectory()).path;
   }
 }
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -61,8 +61,7 @@ class CoreStoreDirectory {
     );
 
     await File(libraryDirectoryDatabaseFilePath).create(recursive: true);
-    await databaseFileToMove.copy(libraryDirectoryDatabaseFilePath);
-    await databaseFileToMove.delete();
+    await databaseFileToMove.rename(libraryDirectoryDatabaseFilePath);
   }
 
   Future<String> getTempDirectory() async {
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -20,7 +20,7 @@ class CoreStoreDirectory {
   ///
   /// TODO: remove this migration algorithm after at least two major releases
   ///
-  /// the release when this migration added is `3.1.14`
+  /// the release when this migration added is `3.1.4`
   Future<void> _migrateDBFileToLibraryDirectory() async {
     final dbFile = await _getDBFileIfExistsInAppDocDir();
 
--- a/packages/flutter/lib/src/storage/core_store_directory_io.dart
+++ b/packages/flutter/lib/src/storage/core_store_directory_io.dart
@@ -1,17 +1,71 @@
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:path_provider/path_provider.dart' as path_provider;
+import 'package:path/path.dart' as path;
 
 class CoreStoreDirectory {
   Future<String> getDatabaseDirectory() async {
-    if (Platform.isIOS) {
+    if (defaultTargetPlatform == TargetPlatform.iOS) {
+      await _migrateDBFileToLibraryDirectory();
+
       return (await path_provider.getLibraryDirectory()).path;
-    } else {
-      return (await path_provider.getApplicationDocumentsDirectory()).path;
+    }
+
+    return (await path_provider.getApplicationDocumentsDirectory()).path;
+  }
+
+  /// Why do we need this migration on iOS? see:
+  /// https://github.com/parse-community/Parse-SDK-Flutter/issues/791
+  ///
+  /// TODO: remove this migration algorithm after at least two major releases
+  ///
+  /// the release when this migration added is `3.1.14`
+  Future<void> _migrateDBFileToLibraryDirectory() async {
+    final dbFile = await _getDBFileIfExistsInAppDocDir();
+
+    if (dbFile != null) {
+      await _moveDatabaseFileToLibraryDirectory(dbFile);
     }
   }
 
-  Future<String?> getTempDirectory() async {
+  Future<File?> _getDBFileIfExistsInAppDocDir() async {
+    final appDocDirPath =
+        (await path_provider.getApplicationDocumentsDirectory()).path;
+
+    final databaseFilePath = path.join(
+      appDocDirPath,
+      'parse',
+      'parse.db',
+    );
+
+    final dbFile = File(databaseFilePath);
+
+    if (await dbFile.exists()) {
+      return dbFile;
+    }
+
+    return null;
+  }
+
+  Future<void> _moveDatabaseFileToLibraryDirectory(
+    File databaseFileToMove,
+  ) async {
+    final libraryDirectoryPath =
+        (await path_provider.getLibraryDirectory()).path;
+
+    final libraryDirectoryDatabaseFilePath = path.join(
+      libraryDirectoryPath,
+      'parse',
+      'parse.db',
+    );
+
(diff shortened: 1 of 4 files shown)


Nidal-Bakir:
@mtrezza @mbfakourii Friendly ping about this PR

mbfakourii:
> @mtrezza @mbfakourii Friendly ping about this PR

Unfortunately, I don't have a Mac system to check, but the code has no logic problems and looks good

mbfakourii (approved):
> @mtrezza @mbfakourii Friendly ping about this PR

Unfortunately, I don't have a Mac system to check, but the code has no logic problems and looks good

mtrezza (changes requested):
See https://github.com/parse-community/Parse-SDK-Flutter/issues/791#issuecomment-1445497694

packages/flutter/lib/src/storage/core_store_directory_io.dart
```diff
@@ -1,11 +1,71 @@
...
+        (await path_provider.getLibraryDirectory()).path;
+
+    final libraryDirectoryDatabaseFilePath = path.join(
+      libraryDirectoryPath,
+      'parse',
+      'parse.db',
+    );
+
+    await File(libraryDirectoryDatabaseFilePath).create(recursive: true);
+    await databaseFileToMove.copy(libraryDirectoryDatabaseFilePath);
```
mtrezza, line 64: Instead of copy + delete, is there a move command? In some file system that simply changes the file pointer instead of actually duplicating the file and then deleting the original.

Nidal-Bakir: Thanks for pointing this out. 
Yes, there is [rename()](https://api.dart.dev/stable/2.19.2/dart-io/File/rename.html) function that will do the trick

Nidal-Bakir:
@mtrezza 
Before we merge this, I have some concerns about the `dbDirectory`function

https://github.com/parse-community/Parse-SDK-Flutter/blob/a19ea872ef31930c5160000eba895ebc26c361b6/packages/flutter/lib/parse_server_sdk.dart#L138-L142

`$dbDirectory/parse`
the `/` here will cause a bug on Windows OS right ? or I'm missing something....
because the resulting path on windows OS will be like this: `dbDirectory/parse\parse.db `
the first `/` should be `\` right? 

so we need to use the join for all the parameters like:
```dart
    final dbDirectory = path.join(
      dbDirectory,
      'parse',
      'parse.db',
    );
```

What do you think?

mtrezza:
Is `CoreStoreDirectory().getDatabaseDirectory();` OS agnostic?

Nidal-Bakir:
@mtrezza 

>Is CoreStoreDirectory().getDatabaseDirectory(); OS agnostic?

In our case yes it is because we use `ApplicationDocumentsDirectory` for any OS other than iOS which uses `LibraryDirectory`

>A new issue would be great. I think we also need to look at this bug separately and I think it merits a separate changelog entry and release

Got it you are right


> It may even be a breaking change for Windows systems if the fix creates a different path for the file

I do not think so! an exception will be thrown I think. I will try it tomorrow and see what will happen.

mtrezza:
I've modified the PR title, could you please review?

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,266 @@
...
+        final dbFileSizeBefore = dbFile.lengthSync();
+        final dbFileLastModifiedBefore = dbFile.lastModifiedSync();
+
+        for (final platform in targetPlatforms) {
+          debugDefaultTargetPlatformOverride = platform;
+
+          // act
+          final dbDirectory = await coreStoreDirectory.getDatabaseDirectory();
+
+          // assert
```
mtrezza, line 168:
```suggestion
          final dbDirectory = await coreStoreDirectory.getDatabaseDirectory();

```

mtrezza (changes requested):
Minor cleanup

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+            dbFileSizeBefore,
+            equals(dbFileSizeAfter),
+            reason: 'the db file should be the same',
+          );
+          expect(
+            dbFileLastModifiedBefore.compareTo(dbFileLastModifiedSyncAfter),
+            equals(0), // 0 if this DateTime [isAtSameMomentAs] [other]
+            reason: 'last modified date should not change',
+          );
+
```
mtrezza, line 161:
```suggestion
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+        for (final platform in targetPlatforms) {
+          debugDefaultTargetPlatformOverride = platform;
+
+          final dbDirectory = await coreStoreDirectory.getDatabaseDirectory();
+          expect(dbFile.existsSync(), isTrue);
+
+          final dbFileSizeAfter = dbFile.lengthSync();
+          final dbFileLastModifiedSyncAfter = dbFile.lastModifiedSync();
+
```
mtrezza, line 150:
```suggestion
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+        final dbFileInNewPath = create1MBParseDBFileInLibraryPath();
+        final dbFileSizeBefore = dbFileInNewPath.lengthSync();
+        final dbFileLastModifiedBefore = dbFileInNewPath.lastModifiedSync();
+        final dbDirectory = await coreStoreDirectory.getDatabaseDirectory();
+        expect(dbFileInNewPath.existsSync(), isTrue);
+
+        final dbFileSizeAfter = dbFileInNewPath.lengthSync();
+        final dbFileLastModifiedAfter = dbFileInNewPath.lastModifiedSync();
+
```
mtrezza, line 112:
```suggestion
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+              'dbDirectory should be the new db dir path for iOS (LibraryDir)',
+        );
+
+        final newDBFilePath = path.join(
+          dbDirectory,
+          'parse',
+          'parse.db',
+        );
+        final newDBFile = File(newDBFilePath);
+
```
mtrezza, line 88:
```suggestion
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+        final oldDBFile = create1MBParseDBFileInAppDocDir();
+        final oldDBFileSize = oldDBFile.lengthSync();
+        final dbDirectory = await coreStoreDirectory.getDatabaseDirectory();
+        expect(
+          oldDBFile.existsSync(),
+          isFalse,
+          reason: 'the old db file should be deleted from app doc dir',
+        );
+
```
mtrezza, line 74:
```suggestion
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+      final result = await path_provider.getApplicationDocumentsDirectory();
+      expect(result.path, applicationDocumentsPath);
+    });
+    test('defaultTargetPlatform should equals iOS', () async {
+      debugDefaultTargetPlatformOverride = TargetPlatform.iOS;
+
+      final platform = defaultTargetPlatform;
+      expect(platform, equals(TargetPlatform.iOS));
+    });
+    test('getTempDirectory() should return kTemporaryPath', () async {
```
mtrezza, line 43:
```suggestion
    
    test('getTempDirectory() should return kTemporaryPath', () async {
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+    test('getLibraryDirectory', () async {
+      final result = await path_provider.getLibraryDirectory();
+      expect(result.path, libraryPath);
+    });
+
+    test('getApplicationDocumentsDirectory', () async {
+      final result = await path_provider.getApplicationDocumentsDirectory();
+      expect(result.path, applicationDocumentsPath);
+    });
+    test('defaultTargetPlatform should equals iOS', () async {
```
mtrezza, line 37:
```suggestion
    
    test('defaultTargetPlatform should equals iOS', () async {
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+    setUp(() {
+      PathProviderPlatform.instance = FakePathProviderPlatform();
+      coreStoreDirectory = CoreStoreDirectory();
+    });
+
+    test('getTemporaryDirectory', () async {
+      final result = await path_provider.getTemporaryDirectory();
+      expect(result.path, kTemporaryPath);
+    });
+    test('getLibraryDirectory', () async {
```
mtrezza, line 28:
```suggestion
    
    test('getLibraryDirectory', () async {
```

packages/flutter/test/src/storage/core_store_directory_io_test.dart
```diff
@@ -0,0 +1,243 @@
...
+          reason: 'last modified date should not change',
+        );
+        expect(
+          dbDirectory,
+          equals(libraryPath),
+          reason:
+              'dbDirectory should be the new db dir path for iOS (LibraryDir)',
+        );
+      });
+      test(
```
mtrezza, line 130:
```suggestion
      
      test(
```

Nidal-Bakir:
@mtrezza 
Thanks for the edits.

LGTM

Nidal-Bakir:
@mtrezza 


> @mtrezza Before we merge this, I have some concerns about the `dbDirectory`function
> 
> https://github.com/parse-community/Parse-SDK-Flutter/blob/a19ea872ef31930c5160000eba895ebc26c361b6/packages/flutter/lib/parse_server_sdk.dart#L138-L142
> 
> `$dbDirectory/parse` the `/` here will cause a bug on Windows OS right ? or I'm missing something.... because the resulting path on windows OS will be like this: `dbDirectory/parse\parse.db ` the first `/` should be `\` right?
> 
> so we need to use the `join` function for all the parameters like:
> 
> ```dart
> Future<String> dbDirectory() async {
>   final dbDirectory = await CoreStoreDirectory().getDatabaseDirectory();
>   return path.join(
>     dbDirectory,
>     'parse',
>     'parse.db',
>   );
> }
> ```
> 
> What do you think?
> 
> Should I fix this in this PR or open a new issue with a new PR for it?
---
I test the path on `windows 10` 



It looks like `windows 10` will understand the path and will work normally and the file was created without any issues. **But** this is not a normal path for windows OS. I do not know what the consequences of doing that.
as you can see this is not normal at all:  `c:\nidal\IdeaProjects\dart_application_1/parse\parse.db`

mtrezza:
Thanks for the lint fix.

> But this is not a normal path for Windows OS.

I think what's going on here is that while historically in MS-DOS it was not a directory operator, at some point (probably many years back) the path interpreter has been updated to accept forward slashes as well. Hence the path you mention is correctly interpreted. I tried this in MS-DOS 6.22 and it's interpreted as a command switch as expected. If you try it in a Windows command prompt and type `cd a/b/c` or cd `a\b/c` then it may work. Another theory is that there is something before the OS on Flutter's side that is correcting the path as an intermediary.

mtrezza (approved):
Looks good!

PR #827, opened 2023-02-25
test: Add tests for `ParseObject` class

Nidal-Bakir:

### Issue Description
Add tests for ParseObject class

### TODOs before merging
Do not merge until all tests are done!

- [x] Add tests
- [ ] A changelog entry


mbfakourii (dismissed):
Thanks for the test, it looks good

packages/dart/test/src/objects/parse_object_test.dart
```diff
@@ -4,7 +4,8 @@ import 'package:mockito/annotations.dart';
 import 'package:mockito/mockito.dart';
 import 'package:parse_server_sdk/parse_server_sdk.dart';
 import 'package:test/test.dart';
-import 'parse_query_test.mocks.dart';
+
+import '../../parse_query_test.mocks.dart';
```
mtrezza, line 8: Why is this change needed if the PR doesn't add a new file?

Nidal-Bakir: Because I moved the test file to a new location
from `test/parse_object_test.dart` to `test/src/objects/parse_object_test.dart`

Its recommended to have your tests folder structure follow the lib folder structure.

In this case, I moved the` parse_object_test.dart` file to a location that mirrors its location in the lib folder.

mbfakourii: > Because I moved the test file to a new location from `test/parse_object_test.dart` to `test/src/objects/parse_object_test.dart`
> 
> Its recommended to have your tests folder structure follow the lib folder structure.
> 
> In this case, I moved the` parse_object_test.dart` file to a location that mirrors its location in the lib folder.

@mtrezza 
I think it's good because the test section is getting bigger

packages/dart/test/src/objects/parse_object_test.dart
```diff
@@ -420,5 +420,127 @@ void main() {
         verifyNoMoreInteractions(client);
       });
     });
+
+    group('update()', () {
```
mbfakourii, line 424: Why is it in the group in this section? I don't think it is necessary

mbfakourii (changes requested):

packages/dart/test/src/objects/parse_object_test.dart
Nidal-Bakir: Every function in the ParseObject will have a group test. so that it will be easier to run tests for a specific function plus its more organamic I think.

mbfakourii: I got it right

mbfakourii (approved):

mtrezza:
Is this ready for merge?

mbfakourii (approved):
There is no case, I think it is ready for merge

Nidal-Bakir:
@mtrezza @mbfakourii 
I plan to write tests for all the functions in this class `ParseObject`.

So currently it's not ready to be merged.

mtrezza:
Thanks for this, I'll convert this to draft for now.

mbfakourii (approved):

mtrezza:
That's an impressive tour de force! 🤯🙌

mbfakourii:
It's great ❤️
just not too many lines? Do you think it is possible to separate?

Nidal-Bakir:
Sure, do you have any preferences?


But in my humble opinion, there is no need to split it because:
* All the tests are about the same class (ParseObject)
* All the tests are testing the logic in that class no more no less.
* Everything is organized in groups 

It's a lot of tests because the ParseObect class is doing a lot of things. plus every function in the test suite lives in a group so it's easy to add/edit/remove any tests for specific functionality.

Nidal-Bakir:
I'm very careful with these tests because we will rely on them to spot any bugs when fixing/adding features to the SDK.

mtrezza:
I haven't looked at how the tests in this PR are grouped, but you may want to take a look at the Parse Server or Parse JS SDK tests to see how they are structured. They surely are no perfect as they amassed over time, but there is a certain grouping, and sometimes even sub-groups within groups, to keep it somewhat organized.

mtrezza:
Just one aspect where I would suggest grouping within 1 file rather than splitting into multiple files is code repetition.

Nidal-Bakir:
Thanks for the suggestion. I did what you suggest.
I think we are good now

Nidal-Bakir:
With that, I have added tests for all the functions in the ParseObject class except the `save()` function I did not add tests for it because there is a lot of thing going on in it, like a lot of things!!

I will add tests for the `save()` function in a new PR so we can talk about it and I can wrap my head around it.

packages/dart/test/src/objects/parse_object/parse_object_array_test.dart
```diff
@@ -0,0 +1,384 @@
+@Skip('get(keyArray) will return _Map<String, dynamic>'
```
mbfakourii, line 1: Does this todo need to be written here?

packages/dart/test/src/objects/parse_object/parse_object_increment_decrement_test.dart
```diff
@@ -0,0 +1,222 @@
+@Skip('get(key) will return _Map<String, dynamic>'
```
mbfakourii, line 1: Too

packages/dart/test/src/objects/parse_object/parse_object_relation_test.dart
```diff
@@ -0,0 +1,133 @@
...
+        // arrange
+        dietPlansObject.removeRelation('usersRelation', [user1, user2]);
+
+        // assert
+        expect(
+          () => dietPlansObject.getRelation('usersRelation'),
+          returnsNormally,
+        );
+      },
+      skip: 'getRelation() will throw Unhandled exception:'
```
mbfakourii, line 113: +

packages/dart/test/src/objects/parse_object/test_utils.dart
```diff
@@ -0,0 +1,107 @@
+import 'package:parse_server_sdk/parse_server_sdk.dart';
```
mbfakourii, line 1: Isn't it better to put this file somewhere else?

mbfakourii (changes requested):
Thanks for all the tests ParseObject. I checked it, it's good in general, only a few questions came up that I asked in the comments

packages/dart/test/src/objects/parse_object/parse_object_array_test.dart
Nidal-Bakir: The `TODO` reminds us that this file is skipped from the tests.

packages/dart/test/src/objects/parse_object/test_utils.dart
Nidal-Bakir: How about we put it in the root `test/ `so any util function can be written in it

What do you think?

mbfakourii: Yes, it is a good suggestion

packages/dart/test/src/objects/parse_object/parse_object_array_test.dart
mbfakourii: Yes, I got it

mtrezza:
Feel free to request another review from `@parse-community/android-sdk-review` when this is ready :-)

Nidal-Bakir:
Yes indeed! it is so helpful..

Okay all the checks are completed and successful
Ready to be merged

Nidal-Bakir:
I'm sorry but I do not think I can add tests for the save() function. I spent all the night trying to understand what is the logic and the flow behind it and couldn't... 

The only way to fix this old, recursive full of if-else-if-else code is to rewrite it from scratch or we need to call the developer who wrote this code and ask them to refactor it.

If you guys have any suggestions, please let me.

mbfakourii (dismissed):
I checked it, I think it's very good, thanks for the tests

mtrezza:
> I'm sorry but I do not think I can add tests for the save() function.

What exactly would you want to test? I assume testing save could simply be done by saving an object and then querying it?

Nidal-Bakir:
You are right, but that is just a part of the story.
we need to account/test:

- every possible variation of composite objects nested in each other.
- Files 
- ACL 
-  Locations 
- Lists 
- nested saved object 
- nested unsaved object
- cycles where two or more unsaved objects depend on each other
- cycles where two or more saved and unsaved objects depend on each other
- relations
- unsaved objects in relation
- any edge case that I do not know about yet.

This is a core function in the SDK, currently, we can not touch it because we might accidentally break things if anyone tries to fix something in this function.

That is why I need to really understand what that function doing.

I will give it another try I'm going to read documentation and code from android and IOS SDKs.

mtrezza:
Why do we need to understand it internally to test it? A basic test inherently treats this as a black box, not caring how it does something but ensuing that the input generates an expected output.

Nidal-Bakir:
> A basic test inherently treats this as a black box

These types of tests are expected when dealing with libraries and packages that are outside our control and we can not test their internal code.

As I said in the previous comment, I will try again but this time I will read code from the 3 SDKs flutter, android, and iOS

If I could not get it in my head, then I will write a black box test. 
 
plus, if there is a bug in this function, which I have experienced in the past month or so. where the recursive function never stops and the dart language will throw StackOverflow but that is a very specific edge case I forage about.

My point of this is to understand every code/function/class in the SDK so I can fix and improve the features

pastordee:
> > A basic test inherently treats this as a black box
> 
> These types of tests are expected when dealing with libraries and packages that are outside our control and we can not test their internal code.
> 
> As I said in the previous comment, I will try again but this time I will read code from the 3 SDKs flutter, android, and iOS
> 
> If I could not get it in my head, then I will write a black box test.
> 
> plus, if there is a bug in this function, which I have experienced in the past month or so. where the recursive function never stops and the dart language will throw StackOverflow but that is a very specific edge case I forage about.
> 
> My point of this is to understand every code/function/class in the SDK so I can fix and improve the features

I think you are 100% right, if you are going to do something, give it your best and in programming it is always good to have a solid understanding of everything. 
PS: take your time.

mtrezza:
Let me clarify what I meant by that. 

We (codecov) are tracking the horizontal test coverage, i.e. how many functions are tested. The vertical coverage is not tracked and we have no means to measure it. Vertical here means also considering internal states that are not exposed to the developer, which requires to understand the internals of the SDK.

While vertical coverage can be important, the most efficient way to increase coverage is to add black box test. That means we improve the SDK stability for developers the fastest.

It's not a question of one or the other. Vertical tests built upon horizontal tests, so over time the test can additionally consider SDK-internal states to make the test more robust.

Nidal-Bakir:
Now I see your point, thanks for making this clear.

I will work on that.

Nidal-Bakir:
I will change this PR to a draft until I include the necessary tests for the `save()` function.

Nidal-Bakir:
@mtrezza 
Should we merge this before or after #852?

mtrezza:
What do you suggest

Nidal-Bakir:
Let's wait, I don't wont to mess up his work with conflicts. I can deal with the conflicts after his work is done.

Nidal-Bakir:
@mtrezza

mtrezza (changes requested):
This PR still contains open discussions on review comments. Please resolve all for the PR to get merged.

Nidal-Bakir:
done

mtrezza (approved):
Looks good!

Issue #835, opened 2023-03-01
Flutter package publishing fails


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description

Publishing the flutter package fails with error:

```
Package validation found the following potential issues:
* The name of "lib/parse_server_sdk.dart", "parse_server_sdk", should match the name of the package, "parse_server_sdk_flutter".
  This helps users know what library to import.
* line 8, column 1 of test/src/storage/core_store_directory_io_test.dart: This package does not have path_provider_platform_interface in the `dependencies` or `dev_dependencies` section of `pubspec.yaml`.
    ╷
  8 │ import 'package:path_provider_platform_interface/path_provider_platform_interface.dart';
    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ╵
* line 9, column 1 of test/src/storage/core_store_directory_io_test.dart: This package does not have plugin_platform_interface in the `dependencies` or `dev_dependencies` section of `pubspec.yaml`.
    ╷
  9 │ import 'package:plugin_platform_interface/plugin_platform_interface.dart';
    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
### Steps to reproduce

See https://github.com/parse-community/Parse-SDK-Flutter/actions/runs/4304405195/jobs/7505394193

### Actual Outcome

Package publish fails.

The CI passed in https://github.com/parse-community/Parse-SDK-Flutter/pull/826, so this is rather unexpected. Maybe we should add the publish dry-run to the PR CI to prevent his in the future?

### Expected Outcome

Package should publish.

[...]

PR #837, opened 2023-03-02
ci: Fix failing CI due to missing platform interface dependencies

Nidal-Bakir:

### Issue Description

Closes: #835

### TODOs before merging

- [ ] Add tests
- [ ] A changelog entry

### DIFF
--- a/packages/flutter/pubspec.yaml
+++ b/packages/flutter/pubspec.yaml
@@ -36,3 +36,4 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   mockito: ^5.2.0
+  path_provider_platform_interface: ^2.0.6
--- a/packages/flutter/pubspec.yaml
+++ b/packages/flutter/pubspec.yaml
@@ -37,3 +37,4 @@ dev_dependencies:
     sdk: flutter
   mockito: ^5.2.0
   path_provider_platform_interface: ^2.0.6
+  plugin_platform_interface: ^2.1.4
--- a/packages/dart/pubspec.yaml
+++ b/packages/dart/pubspec.yaml
@@ -30,3 +30,4 @@ dev_dependencies:
   build_runner: ^2.1.11
   mockito: ^5.2.0
   test: ^1.21.4
+  path_provider_platform_interface: ^2.0.6
--- a/packages/dart/pubspec.yaml
+++ b/packages/dart/pubspec.yaml
@@ -30,4 +30,3 @@ dev_dependencies:
   build_runner: ^2.1.11
   mockito: ^5.2.0
   test: ^1.21.4
-  path_provider_platform_interface: ^2.0.6


packages/flutter/pubspec.yaml
```diff
@@ -36,3 +36,4 @@ dev_dependencies:
   flutter_test:
     sdk: flutter
   mockito: ^5.2.0
+  path_provider_platform_interface: ^2.0.6
```
mbfakourii, line 39: Welcome to the Parse Flutter SDK review team ❤️
and i tested the code with this package with the following command
```dart
flutter pub publish --dry-run
```
One of the errors has been solved, but the other error remains, I suggest you add the following package
```
plugin_platform_interface: ^2.1.4
```

mbfakourii (changes requested):

packages/flutter/pubspec.yaml
Nidal-Bakir: it's working now.

mbfakourii (approved):

mtrezza:
Is this a bug that affects only our CI or also the published SDK for developers? If the latter, what is the bug that can occur for developers?

mbfakourii:
> Is this a bug that affects only our CI or also the published SDK for developers? If the latter, what is the bug that can occur for developers?

I tested the version of ```parse_server_sdk_flutter: ^3.1.3```, there is no problem on the developer side, probably there is a problem in CI.

mtrezza:
Could it be that issues only occur on specific platforms? Looking at the docs of the [path_provider_platform_interface](https://pub.dev/packages/path_provider_platform_interface):
> This interface allows platform-specific implementations of the path_provider plugin, as well as the plugin itself, to ensure they are supporting the same interface.

Nidal-Bakir:
I think it's normal.
Because the **Parse_SDK** is a **package** and not a **plugin**  if you look at the [pubspec.yaml]( https://github.com/flutter/packages/blob/4aedb859de141673f2bd9985bb46eceac407d9be/packages/path_provider/path_provider/pubspec.yaml#L41) in `path_provider` you will see `plugin_platform_interface: ^2.0.0` under the dev_dependencies and dependencies that's because it needed in tests and in production.

We need these plugins in tests so it's safe to put them under dev_dependencies and the users of the package will not be affected.

Because `pub` will NOT take into account the dev_dependencies from our pubspec.yaml
From the [docs](https://dart.dev/tools/pub/dependencies#dev-dependencies) 
>Dev dependencies differ from regular dependencies in that dev dependencies of packages you depend on are ignored

So do we need new releases? No there is no need for it.

Do we need a better way to test using `path_provider`? Yes
we can do that using DI (Dependency Injection) and not call the `path_provider` functions 
```dart
// core_store_directory_io.dart

class CoreStoreDirectory {
  final PathProvider _pathProvider;

  CoreStoreDirectory(this._pathProvider);

    return (await _pathProvider.getLibraryDirectory()).path;
  }

  Future<String> getTempDirectory() async {
    return (await _pathProvider.getTemporaryDirectory()).path;
  }



//--> path_provider.dart

 abstract class PathProvider {
  Future<Directory> getTemporaryDirectory();
  Future<Directory> getLibraryDirectory();
  Future<Directory> getApplicationDocumentsDirectory();
}

class PathProviderImpl extends PathProvider {
  @override
  Future<Directory> getApplicationDocumentsDirectory() {
    return path_provider_package.getApplicationDocumentsDirectory();
  }

  @override
  Future<Directory> getLibraryDirectory() {
    return path_provider_package.getLibraryDirectory();
  }

  @override
  Future<Directory> getTemporaryDirectory() {
    return path_provider_package.getTemporaryDirectory();
  }
}

```

mtrezza:
Ok, so can we merge this with commit type `ci`?

mtrezza (approved):
Ok, so let's merge this with commit type `ci`, meaning no new release.

Issue #839, opened 2023-03-02
Check for `dart pub publish --dry-run` errors in CI


Nidal-Bakir:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
The current CI does not account for `dart pub publish` errors

### Steps to reproduce
See https://github.com/parse-community/Parse-SDK-Flutter/actions/runs/4304405195/jobs/7505394193
Our last publish fails due to an error from `dart pub publish`. so we need a way to check for these errors in the `ci.yml` file so we can spot them immediately.

### Actual Outcome
Error when trying to publish to pub.dev

### Expected Outcome
The error should be spotted before the publishing stage.

PR #840, opened 2023-03-02
ci: Add publish dry-run check for dart package

Nidal-Bakir:

### Issue Description

Closes: #839

### Approach

start a job to check for any errors when running `dart pub publish --dry-run` 

### TODOs before merging
All set

- [ ] Add tests
- [ ] A changelog entry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,18 +76,6 @@ jobs:
         uses: dart-lang/setup-dart@v1
       - name: publish dry run
         run: cd packages/dart && dart pub publish --dry-run
-  check-publish-dry-run-flutter:
-    name: publish dry-run (flutter)
-    timeout-minutes: 5
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-      - name: Setup dart
-        uses: dart-lang/setup-dart@v1
-      - name: publish dry run
-        run: |
-          (cd packages/flutter && dart pub publish --dry-run 1> flutter_dry_run_stdout 2> flutter_dry_run_stderr) 
   check-flutter:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
         uses: dart-lang/setup-dart@v1
       - name: publish dry run
         run: |
-          (cd packages/flutter && dart pub publish --dry-run) 
+          (cd packages/flutter && dart pub publish --dry-run 1> flutter_dry_run_stdout 2> flutter_dry_run_stderr)
   check-flutter:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,6 @@ jobs:
           cd packages/flutter && dart pub publish --dry-run 1> flutter_dry_run_stdout 2> flutter_dry_run_stderr
           DRY_RUN_STD_ERR_NO_SPACES=$(tr -d "[:space:]" < flutter_dry_run_stderr)
           cat flutter_dry_run_stdout flutter_dry_run_stderr
-          [[ $DRY_RUN_STD_ERR_NO_SPACES == $CURR_ERR ]] && echo "no errors" || echo "error from dart pub publish --dry-run" && exit 1
   check-flutter:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


Nidal-Bakir:
@mtrezza 
Do you know why I can not redirect the stdout/stderr to a file in GitHub action?

`(dart pub publish --dry-run 1> flutter_dry_run_stdout 2> flutter_dry_run_stderr)`

mtrezza:
What is this error? Should we rename a package?

Nidal-Bakir:
@mtrezza 
We just need to rename the file in `Parse-SDK-Flutter/packages/flutter/lib` 
from `parse_server_sdk.dart`  to `parse_server_sdk_flutter.dart`

Note that anyone is importing the file using the old name will get an error but its fine nothing to worry about

mtrezza:
So this change won't break the app at runtime because the developer will always use an IDE and the compiler will always throw an error?

Nidal-Bakir:
> this won't be a breaking change for existing apps
No. any user of the package will get compile time error.

But it is annoying because let's say you have 15 files and in each one of them you are importing the old file name. in this case you need to fix each one of them using the new file name import.

mtrezza:
I think then formally it is a breaking change, as it requires a migration, no matter how easy the migration is. Nothing wrong with that - we have a "Breaking Change" section in our changelog style where we can add details about how to migrate. The major version increment will also highlight to users that a manual migration needs to be done.

Nidal-Bakir:
Got it, in this case, if we will introduce a breaking change let's make it count. I mean introducing a breaking change just because we changed the way the package is been imported is not enough of a reason I think.

>How about we merge the renaming PR first, then this PR with dart + flutter dry-run?

I do not think it's a good idea because the package users already using the import as it is. and it is not a big deal. I suggest merging this PR with a dry-run for dart only for now. 

And after that, we will open a new 2 PRs:
* renaming the file
* adding a dry-run for flutter 

and not merge them right away we wait and see if any bug or feature or other reason requires a breaking change then we merge these 2 PRs with the newly introduced breaking change if any, if no breaking change is introduced for example 1 or 2 months then we can merge this 2 PRs.

> Since we'll do a major version release, we can also add other breaking changes (in separate PRs) if you can think of any. Just for efficiency as we try to accumulate breaking changes into a single major release. That is to not surprise developers with breaking releases too often.

You are right. so as I stated earlier let us wait about 1 to 2 months and decide then what to do. 

What do you think?

mtrezza:
That sounds good. Let me try to make the pending flutter release that failed last time. It should be fixed with https://github.com/parse-community/Parse-SDK-Flutter/issues/837 right?

Nidal-Bakir:
Yes, we fix it.

mtrezza:
I tried to release flutter, but it failed - obviously - because it's also doing the dry-run before the actual publishing to ensure all goes well. It now only fails for the reason of the naming issue.

Nidal-Bakir:
how this error did not prevent older releases?
I thought it is just a warning

mtrezza:
Hah, good question!

Nidal-Bakir:
I do not recommend option (b). Forcibly pushing the package to pub.dev is not a good idea `dry-run` is there for a good reason.

on the other hand, If we choose option (a), we won't be able to release bug fixes until we decided it's time to fix this and introduce a breaking change.

I think option (a) is the more suitable and let's wait for a week or two and introduce a breaking change with any bug fixes if any.

mtrezza:
> we won't be able to release bug fixes until we decided it's time to fix this and introduce a breaking change.

How so? We can release bug fixes any time after the release of Flutter 4.0.0

Nidal-Bakir:
> My thinking is, it would be nice to wait with (a) until we have more stuff for a breaking change release

I thought that we will wait a bit before we release flutter 4.0.0

Nidal-Bakir:
We can do it now if you do not have any issues with it

mtrezza:
Sure we can wait, why would you wait?

mtrezza:
I follow you on this, what do you suggest, wait or release?

Nidal-Bakir:
In situations like this, it would be wise to wait a few days. We do not want to rush anything.

Tomorrow or the day after tomorrow, I will open a new issue and PR to introduce the breaking change for the flutter package.

mbfakourii:
> That sounds good. @mbfakourii what do you think?
> 
> Let me try to make the pending flutter release that failed last time. It should be fixed with #837 right?

Yes, I think it's good too

mtrezza:
Should we merge this PR already as dart-only or after the flutter major release with dart + flutter?

Nidal-Bakir:
Yes, let's merage this as add dry-run for the dart package. 

I will open a new 2 PRs:

 *   renaming the file so we fix the wrong import for the flutter package
 *   adding a dry-run for flutter

mtrezza (approved):
Looks good!

Issue #850, opened 2023-03-08
Merge flutter and dart package


mbfakourii:
### New Feature / Enhancement Checklist
- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Current Limitation
I have a question, why do we have two packages? Is it possible to integrate Flutter package with Dart? I checked that there are only 6 files in the Flutter version and most of them are utilities!
With this, we will no longer get involved in the conflicts of two packages.

### Feature / Enhancement Description
I suggest to merge two packages into one package

### Tasks

- [ ] Add migration guide to new package

### Roadmap

- 1. Merge https://github.com/parse-community/Parse-SDK-Flutter/pull/855

mbfakourii:
The only basic part that exists is ParseLiveList, which can be implemented in Flutter in different ways.

I think it is possible to prepare an example for this issue instead of having a separate package!

mtrezza:
Maybe @phillwiggins who is the original author of this SDK can say something to it. I believed it started with the dart package and flutter was developed later on, on top of it.

Nidal-Bakir:
I do not have any answers to this question.

But at this point, it is nearly impossible to merge them together with thousands of developers using each one.

mtrezza:
I don't see an issue with merging them and releasing as a new version, maybe with a transitory phase for developers to migrate.

Nidal-Bakir:
One package is for Dart and another one is built on top of the Dart package, for Flutter.

The reason for this is that Flutter has its own set of unique widgets and libraries that are not available in Dart. When developers create packages that are specifically designed for Flutter, they need to use these Flutter widgets and libraries to make their package work properly within the Flutter framework.

So, the Dart package contains the core logic and functionality of the library, while the Flutter package contains the code that is necessary to integrate that functionality with Flutter-specific widgets and libraries. This separation of concerns makes it easier for developers to use the package in their Flutter projects, without having to worry about the underlying Dart code.

For example, let's say we need to implement a portable UI like login/sign-in like firebase UI where we will deal with logging in using email and password, using a google account, or using a Twitter account. well, we can do just that in the flutter package. rather than create a separate package to do that.

[...]

PR #852, opened 2023-03-11
fix: Remove flutter version and add `example` and `example_ui`

mbfakourii:

### Issue Description
In order to reduce conflicts, it was decided to remove the Flutter version, so the following tasks should be done

- [x] Remove the Flutter version
- [x] Restructuring the repository
- [x] Create an ```example``` to test the package
- [x] Create an example called ```example_ui``` for graphic widgets
- [x] Transferring the required codes of the Flutter package to the new package
- [x] Transferring the required codes of the Flutter example project to the new example package
- [x] Transferring the required codes of the Flutter example_livelist project to the new example_ui package
- [x] Writing codes related to the ```example``` project
- [x] Writing the primary codes related to the ```example_ui``` project
- [x] Solve CI problems
- [x] Re-examine README.md

Closes: #850

### Approach
n/a

### TODOs before merging
- [x] Add tests
- [ ] A changelog entry


mbfakourii:
Are there other tasks?

Nidal-Bakir:
If we will merge this we need first to publish a new version of the Flutter SDK to inform the package user about this transition.

TODOs for the README file in the Flutter SDK:

- state that this package will no longer receive support and we are transitioned to the Dart SDK
- explain why we are doing that
- explain that the Flutter was just a wrapper around the Dart SDK and this is not a new package
- inform the package users that the SDK API will not change and everything is in its place.

TODOs for the README file in the Dart SDK(new):
- migration guide for the developer coming from the Flutter SDK

let me know if I missed anything.

Nidal-Bakir (approved):

Nidal-Bakir:
>Are there other tasks?
For now, I do not think of any other tasks.

mbfakourii:
Almost all tasks have been completed. please check it and not sure if CI works !

README.md
```diff
@@ -1,4 +1,5 @@
-![parse-repository-header-sdk-flutterdart](https://user-images.githubusercontent.com/5673677/166121476-3838ac9f-a437-409c-bb28-6e1dfcc36b1c.png)
+![parse-repository-header-sdk-dart](https://user-images.githubusercontent.com/5673677/166120960-ea1f58e3-a62b-4770-b541-f64186859339.png)
```
mtrezza, line 1: What should this then be officially called? The Parse Dart SDK or the Parse Flutter SDK? Both names are currently used in this PR.

The repository itself is called parse-sdk-flutter.

README.md
```diff
@@ -13,15 +14,890 @@
 
 ---
 
-These libraries give you access to the powerful Parse Server backend from your Dart/Flutter app. For more information about Parse Platform and its features visit [parseplatform.org](https://parseplatform.org/).
+This library gives you access to the powerful Parse Server backend from your Dart app. For more information on Parse Platform and its features, visit [parseplatform.org](https://parseplatform.org). The Flutter package was moved [here](https://pub.dev/packages/parse_server_sdk_flutter). If you are using Flutter see [this guide](https://github.com/parse-community/Parse-SDK-Flutter/blob/release/2.0.0/docs/migrate-2-0-0.md) for how to migrate.
```
mtrezza, line 17: I believe the migrate 2-0-0 document is deleted in this PR?

Would this release benefit from a new migration document?

.github/workflows/ci.yml
```diff
@@ -23,7 +22,7 @@ jobs:
         with:
           cache: true
       - name: Lint
-        run: flutter format --output=none --set-exit-if-changed packages/flutter
+        run: flutter format --output=none --set-exit-if-changed /
```
Nidal-Bakir, line 25: since now we are using dart not Flutter, any Flutter-related job is not needed anymore.

Nidal-Bakir, line 25: Since now we are using dart not Flutter, any Flutter-related job is not needed anymore.
you can delete `check-lint-flutter` job

.github/workflows/ci.yml
```diff
@@ -34,7 +33,7 @@ jobs:
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
       - name: Lint
-        run: dart format --output=none --set-exit-if-changed packages/dart
+        run: dart format --output=none --set-exit-if-changed /
```
Nidal-Bakir, line 36: dart format --output=none --set-exit-if-changed `./`

.github/workflows/ci.yml
```diff
@@ -47,11 +46,9 @@ jobs:
         with:
           cache: true
       - name: Install dependencies
-        run: |
-          (cd packages/dart && flutter pub get)
-          (cd packages/flutter && flutter pub get)
+        run: flutter pub get
       - name: Analyze code
-        run: flutter analyze packages/flutter --no-fatal-infos
+        run: flutter analyze / --no-fatal-infos
```
Nidal-Bakir, line 51: you can delete this as well `check-code-analysis-flutter`

.github/workflows/ci.yml
```diff
@@ -75,7 +72,7 @@ jobs:
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
       - name: publish dry run
-        run: cd packages/dart && dart pub publish --dry-run
+        run: dart pub publish --dry-run
   check-flutter:
```
Nidal-Bakir, line 76: Delete `check-flutter` job

.github/workflows/ci.yml
```diff
@@ -129,15 +122,15 @@ jobs:
...
         uses: dart-lang/setup-dart@v1
       - name: Install dependencies
-        run: dart pub get --directory packages/dart
+        run: dart pub get
       - name: Run build_runner
-        run: (cd packages/dart && dart run build_runner build --delete-conflicting-outputs)
+        run: dart run build_runner build --delete-conflicting-outputs
       - name: Run tests
-        run: (cd packages/dart && dart test --coverage=coverage)
+        run: dart test --coverage=coverage
```
Nidal-Bakir, line 129:
```bash 
dart test --coverage=./coverage --concurrency=4 --test-randomize-ordering-seed random
```

.github/workflows/ci.yml
```diff
@@ -129,15 +122,15 @@ jobs:
...
       - name: Convert code coverage
         # Needs to be adapted to collect the coverage at all platforms if platform specific code is added.
         if: ${{ always() && matrix.os == 'ubuntu-latest' }}
-        working-directory: packages/dart
+        working-directory: /
         run: |
           dart pub global activate coverage
           dart pub global run coverage:format_coverage -i coverage/test -o coverage/lcov.info --lcov --packages=.dart_tool/package_config.json --report-on=lib
```
Nidal-Bakir, line 136:
```bash
 dart pub global run coverage:format_coverage --packages=.dart_tool/package_config.json --report-on=lib --lcov -o ./coverage/lcov.info -i ./coverage
```

.github/workflows/ci.yml
```diff
@@ -146,7 +139,7 @@ jobs:
         # Needs to be adapted to collect the coverage at all platforms if platform specific code is added.
         if: ${{ always() && matrix.os == 'ubuntu-latest' }}
         with:
-          files: packages/dart/coverage/lcov.info
+          files: coverage/lcov.info
           fail_ci_if_error: true
```
Nidal-Bakir, line 143:
``` yml
      - name: Upload code coverage
        uses: codecov/codecov-action@v3
        # Needs to be adapted to collect the coverage at all platforms if platform-specific code is added.
        if: ${{ always() && matrix.os == 'ubuntu-latest' }}
        with:
          files: ./coverage/lcov.info
          fail_ci_if_error: true
```

.github/workflows/release-automated.yml
```diff
@@ -14,16 +14,6 @@ jobs:
...
-          dryRunOnly: false
-      - name: Publish flutter package
-        if: ${{ startsWith(github.ref_name, 'flutter') }}
-        uses: k-paxian/dart-package-publisher@v1.4
-        with:
-          accessToken: ${{ secrets.PUBDEV_GOOGLE_ACCOUNT_ACCESS_TOKEN  }}
-          refreshToken: ${{ secrets.PUBDEV_GOOGLE_ACCOUNT_REFRESH_TOKEN }}
-          relativePath: packages/flutter
-          flutter: true
+          relativePath: /
```
Nidal-Bakir, line 17:
```yml     
     relativePath: ./
```

.github/workflows/release-manual.yml
```diff
@@ -32,16 +32,6 @@ jobs:
...
-          dryRunOnly: false
-      - name: Publish flutter package
-        if: github.event.inputs.package == 'flutter'
-        uses: k-paxian/dart-package-publisher@v1.4
-        with:
-          accessToken: ${{ secrets.PUBDEV_GOOGLE_ACCOUNT_ACCESS_TOKEN  }}
-          refreshToken: ${{ secrets.PUBDEV_GOOGLE_ACCOUNT_REFRESH_TOKEN }}
-          relativePath: packages/flutter
-          flutter: true
+          relativePath: /
```
Nidal-Bakir, line 35:
```yml     
     relativePath: ./
```

Nidal-Bakir (changes requested):

.github/workflows/ci.yml
```diff
@@ -75,7 +72,7 @@ jobs:
       - name: Setup dart
         uses: dart-lang/setup-dart@v1
```
Nidal-Bakir, line 73: dart-lang/setup-dart@v1.4

Nidal-Bakir (changes requested):

README.md
Nidal-Bakir: well, it should be Parse Dart SDK because all the code in the SDK uses pure Dart.

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 4.0.0
 homepage: https://github.com/parse-community/Parse-SDK-Flutter
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
+  sdk: '>=2.19.2 <3.0.0'
```
Nidal-Bakir, line 7: `sdk: '>=2.19.2 <3.0.0'` 
I do not recommend making the constraints that tight unless you have a good reason.. like language-specific features that you are using and only work if we are on `>=2.19.2 `

if you do not have any language-specific features or other dependence requires `>=2.19.2`
I suggest making the constraints a little bit looser:
`sdk: '>=2.17.0 <3.0.0'`

README.md
mtrezza: Should be create a new repository Parse-SDK-Parse and archive this one?

Nidal-Bakir: I do not know the consequence of renaming the repo. the one I know for sure is that all the people will have a broken upstream remote git-URL

will that create a problem when we try to publish to pub.dev?

mtrezza: Renaming a repo on GitHub won't break the old URL. GitHub maintains a redirect.

Nidal-Bakir: > I just don't know whether the registry has a concept for renaming where is also redirects.

I did not follow you here, can you tell me what a registry is in this context

mbfakourii: Thank you, by the way, it was my question too, I think it would be better to change the repository to a better name.

README.md
mbfakourii: This version is related to Doc Dart, of course, the migration to version 2 section is no longer needed, I will add things for the new version.

.github/workflows/ci.yml
mbfakourii: This job was deleted

.github/workflows/ci.yml
mbfakourii: This job was deleted

.github/workflows/ci.yml
mbfakourii: It was done

.github/workflows/ci.yml
mbfakourii: This job was deleted but this section had many codes, are you sure?

.github/workflows/ci.yml
mbfakourii: It was done

.github/workflows/ci.yml
mbfakourii: It was done

.github/workflows/release-automated.yml
mbfakourii: It was done

.github/workflows/release-manual.yml
mbfakourii: It was done

.github/workflows/ci.yml
mbfakourii: Four places of this topic were used, all four places were changed

pubspec.yaml
mbfakourii: It was done

README.md
mbfakourii: I also agree with ```parse_sdk_dart```

mbfakourii:
I made the suggested changes, is it possible to check again and is it ready for merge?

.github/workflows/ci.yml
```diff
@@ -120,26 +56,26 @@ jobs:
...
         run: dart pub get
       - name: Run build_runner
         run: dart run build_runner build --delete-conflicting-outputs
       - name: Run tests
-        run: dart test --coverage=coverage
+        run: dart test --coverage=./coverage --concurrency=4 --test-randomize-ordering-seed random
       - name: Convert code coverage
         # Needs to be adapted to collect the coverage at all platforms if platform specific code is added.
         if: ${{ always() && matrix.os == 'ubuntu-latest' }}
         working-directory: /
```
Nidal-Bakir, line 69: `working-directory: /` 
remove this line not needed anymore

Nidal-Bakir (changes requested):

.github/workflows/ci.yml
mbfakourii: It was done

Nidal-Bakir (changes requested):
in the root of the project `analysis_options.yaml` add this 
``` yml

analyzer:
  exclude:
    - example/**
    - example_ui/**
```

Nidal-Bakir:
We still need a migration guide for the developer coming from the Flutter SDK

Nidal-Bakir (changes requested):

Nidal-Bakir (changes requested):
Delete the `tools/` folder. not needed anymore. plus the `dart pub publish --dry-run` command will report the folder with the following error:
``` bash 
Package validation found the following potential issue:
* Rename the top-level "tools" directory to "tool".
  The Pub layout convention is to use singular directory names.
  Plural names won't be correctly identified by Pub and other tools.
  See https://dart.dev/tools/pub/package-layout.

Package has 1 warning.

```

mbfakourii:
> Delete the `tools/` folder. not needed anymore. plus the `dart pub publish --dry-run` command will report the folder with the following error:
> 
> ```shell
> Package validation found the following potential issue:
> * Rename the top-level "tools" directory to "tool".
>   The Pub layout convention is to use singular directory names.
>   Plural names won't be correctly identified by Pub and other tools.
>   See https://dart.dev/tools/pub/package-layout.
> 
> Package has 1 warning.
> ```

Deleted

.github/workflows/ci.yml
```diff
@@ -60,11 +29,11 @@ jobs:
...
         uses: actions/checkout@v3
       - name: Setup dart
-        uses: dart-lang/setup-dart@v1
+        uses: dart-lang/setup-dart@v1.4
       - name: Install dependencies
-        run: dart pub get --directory packages/dart
+        run: dart pub get
       - name: Analyze code
-        run: dart analyze packages/dart
+        run: dart analyze
```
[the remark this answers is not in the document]
mbfakourii: > in the root of the project `analysis_options.yaml` add this
> 
> ```yaml
> analyzer:
>   exclude:
>     - example/**
>     - example_ui/**
> ```

added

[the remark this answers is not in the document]
mbfakourii: added

mbfakourii:
> in the root of the project `analysis_options.yaml` add this
> 
> ```yaml
> analyzer:
>   exclude:
>     - example/**
>     - example_ui/**
> ```

added

mbfakourii:
@Nidal-Bakir Thanks ci fixed

Nidal-Bakir:
If you look at the `parse_server_sdk.dart` in the Flutter package and the `parse_server_sdk.dart` in the Dart package. You will see the difference, the flutter package uses default values if the developer did not provide a value

here is a list of the parameters that should be required in the Dart SDK from now on... 

-  appPackageName
- appName
- appVersion
- locale
- coreStore (CoreStoreSharedPrefsImp,CoreStoreMemoryImp,CoreStoreSembastImp)
- connectivityProvider  (For liveQuery)
- fileDirectory
- appResumedStream (For liveQuery)

kindly compare the two files and see the issues that will pop in the face of the developers because the defaults are not there anymore

Nidal-Bakir:
One of the downsides of this transition is that the package user should provide arguments for parameters that were optional ones previously.

catalunha:
I'm following this journey message by message. And the solution is fantastic. Because actually Parse Server is more Dart than Flutter.
But @Nidal-Bakir last approach made me curious.

I use b4a's ParseServer on Dart and I don't need all these parameters. Because with the elimination of Flutter would I need it?

mbfakourii:
> If you look at the `parse_server_sdk.dart` in the Flutter package and the `parse_server_sdk.dart` in the Dart package. You will see the difference, the flutter package uses default values if the developer did not provide a value
> 
> here is a list of the parameters that should be required in the Dart SDK from now on...
> 
> * appPackageName
> * appName
> * appVersion
> * locale
> * coreStore (CoreStoreSharedPrefsImp,CoreStoreMemoryImp,CoreStoreSembastImp)
> * connectivityProvider  (For liveQuery)
> * fileDirectory
> 
> kindly compare the three files:
> 
> * [parse_server_sdk.dart](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/flutter/lib/parse_server_sdk.dart) (Flutter SDK)
> * [ parse_server_sdk.dart](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/dart/lib/parse_server_sdk.dart) (Dart SDK)
> * [parse_core_data.dart](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/packages/dart/lib/src/data/parse_core_data.dart) (Dart SDK)
> 
> and see the issues that will pop in the face of the developers because the defaults are not there anymore

Do you think this description should be in a new file or in README.md?

mbfakourii:
> I'm following this journey message by message. And the solution is fantastic. Because actually Parse Server is more Dart than Flutter. But @Nidal-Bakir last approach made me curious.
> 
> I use b4a's ParseServer on Dart and I don't need all these parameters. Because with the elimination of Flutter would I need it?

These parameters are optional and do not need to be sent to the server, but an explanation can be given about coreStore.

Nidal-Bakir:
> I use b4a's ParseServer on Dart and I don't need all these parameters. Because with the elimination of Flutter would I need it?

If you are using the _Parse Dart SDK_  and the default configuration work with you then you are good to go no need to pass any additional argument.

But the Flutter framework users will have to pass arguments to these parameters . Because the _Parse Dart SDK_ will have different behavior than the _Parse Flutter SDK_ if they didn't pass arguments to these parameters.

For example, the _Parse Flutter SDK_ will automatically provide a `connectivityProvider` to the SDK, but the _Parse Dart SDK_ will not provide `connectivityProvider` and will keep it `null`.

The `liveQuery` feature expect a `connectivityProvider` so when the connection is cut off and then restored again later the `liveQuery` will try to reconnect to the server. if the user did not pass a `connectivityProvider` the liveQuery will **NOT** work
https://github.com/parse-community/Parse-SDK-Flutter/blob/94be0442883d22056a1e898384abc3adf1495990/packages/dart/lib/src/network/parse_live_query.dart#L42-L50

Another example, the _Parse Flutter SDK_ will provide the `CoreStoreSharedPrefsImp` as default storage but the _Parse Dart SDK_ will not provide any storage option and will fallback to the `CoreStoreMemoryImp`

https://github.com/parse-community/Parse-SDK-Flutter/blob/94be0442883d22056a1e898384abc3adf1495990/packages/dart/lib/src/data/parse_core_data.dart#L43

Nidal-Bakir:
We need to make it crystal clear that the `liveQuery` feature will not work without a `connectivityProvider` argument

Nidal-Bakir:
> Do you think this description should be in a new file or in README.md?

We need a proper migration guide, steps, and code documentation (parameters and behaviors)


Parse_Dart_SDK is   **_NOT_** one-drop replacement of Parse_Flutter_SDK

catalunha:
Thanks for the detailed explanation @Nidal-Bakir 
I believe that some users of Parse_Flutter_SDK like me will not have this vision due to the facilities already presented by Parse_Flutter_SDK.
I support your warnings that the README.md should contain a detailed step-by-step migration to Parse_Dart_SDK.
Avoiding affecting users in the continuity of their projects.
Once I have the beta version of this migration it will be a pleasure to test it on some of my projects to help them.
Thanks to you and everyone else for this fantastic contribution.

Nidal-Bakir:
@mtrezza 
Can I help him with adding code, documentation, and guides?
It will be faster if we collaborate.


If yes, do I have a write permission to push commits to this PR?

mtrezza:
@Nidal-Bakir I don't assume that you have write permissions given your team membership, but you could just try to push a commit. If it fails, then @mbfakourii could give you access to his fork so you can push to his branch.

Nidal-Bakir:
To be honest the developer will write all the code that was previously provided by the Flutter SDK. 
which is:

- init a database with a path using path_porvider package if they need to use Sembast database
- init a sharedPreferences object from the sharedPreferences package if they need to use the sharedPreferences as a database
- create a connectivityProvider class and instantiate it if the need to use the liveQuery
- specifying the fileDirectory so the SDK can store the downloaded files again using path_provider package
- specifying appName, appVersion, appPackageName the Flutter SDK was using PackageInfo package

in summary, we are asking the developers to code the Flutter SDK

And that is what we called _boilerplate_.

mbfakourii:
@Nidal-Bakir 
Do you have push access now?

catalunha:
I fully agree @Nidal-Bakir 

Improvement for some would complicate for many.

Considering my level of knowledge and necessary work compared to what I already have to use, it would not be good news.

mbfakourii:
> I fully agree @Nidal-Bakir
> 
> Improvement for some would complicate for many.
> 
> Considering my level of knowledge and necessary work compared to what I already have to use, it would not be good news.

I think it can be solved with an explanation

Nidal-Bakir:
> Do you have push access now?

Yes, thanks

Nidal-Bakir:
let us wait and see what @mtrezza will say.
it's up to him to decide what should we do.

mbfakourii:
@Nidal-Bakir 

The ```coreStore (CoreStoreSharedPrefsImp,CoreStoreMemoryImp,CoreStoreSembastImp) ``` problem has been solved in [this file](https://github.com/parse-community/Parse-SDK-Flutter/blob/23402113d8bd9bfa55ed9599d93e978c84715237/lib/src/storage/core_store_sp_impl.dart). The idea of using dynamics was to see if it is a good method or not.

Nidal-Bakir:
That's what the Flutter SDK was doing!! 
now the developers need to do it manually.

as @catalunha  said:
> Considering my level of knowledge and necessary work compared to what I already have to use, it would not be good news.

The package users already not liking this transition!
because they need to write code that was hidden in the Flutter SDK.

do  not forget all of this:
> 
>     * init a database with a path using path_porvider package if they need to use Sembast database
> 
>     * create a connectivityProvider class and instantiate it if they need to use liveQuery
> 
>     * specifying the fileDirectory using path_provider package so the SDK can store the downloaded files
> 
>     * specifying appName, appVersion, appPackageName the Flutter SDK was using PackageInfo package
> 
> 
> in summary, we are asking the developers to code the Flutter SDK

Nidal-Bakir:
The SDK before this transition was plug-and-play no extra thing you need to do just set your parse API keys and you're good to go.

But now they need to install the SDK set the parse API keys, set up the database storage, file storage, connectivityProvider for the live query, and specify the appName, appVersion, and appPackageName.

Literally, no one will like this over the old way. which was basically a plug-and-play no extra configuration was required.

mtrezza:
> set up the database storage, file storage

Can you help me understand why we could not provide this in the new single SDK?

Nidal-Bakir:
sure, here is the full store...

> Can you help me understand why we could not provide this in the new single SDK?

Those packages depend on Flutter in their dependency to make platform-specific calls. but the new SDK should be os-agnostic. because it's pure dart and if we use these packages in the new SKD now we are not a pure Dart package and we have a platform-specific code, even though the SDK itself not making these calls.

If we need to add these packages in the SDK we need to migrate the other way around. we need to remove the dart SDK and migrate to the Flutter SDK. then we can use packages that make calls to platform-specific functions. because in this case, a (Flutter package) package uses should use this package in the Flutter context (in a Flutter project) and not in a Dart project.

Another unexpected thing that will face the developers if we add these packages to the Dart SDK is that the developers expect the package to be pure dart so they can use it in a CLI for example or any context outside the Flutter framework, but here is the plot twist the SDK will download the entire Flutter framework for this simple console application because our dependency graph contains Flutter Framework itself as a dependency of a dependency that we are using. 

The following code is from the [path_provider](https://github.com/flutter/packages/blob/11aab473cbe5819650eb43c6db9165c6e798de3c/packages/path_provider/path_provider/pubspec.yaml#L25-L27) package:
``` yml 
dependencies:
  flutter:
    sdk: flutter
```

As you can see the path_provider package depends on the Flutter framework.

In the world of Flutter and Dart packages, we have two types:

1.  Dart package (pure Dart): the package contains only dart code no platform-specific calls, nor Flutter-related code. an example of these types of packages would be math, HTTP, Dio, or any other package that is contained in itself. i.e. can run without interaction with the host OS or the Flutter framework.
2. Flutter packages which in turn have tow types:
   * Flutter Package: a package that depends on the Flutter framework to do its work. you can **_use_** any package that calls functions on the host OS like path_provider.
   * Flutter Plugin: it is a type of package that will call the underlying platform by itself like the path_provider. using platform channels for example.

 [Package types](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#types)
> 
> Packages can contain more than one kind of content:
> 
> Dart packages:
>     General packages written in Dart, for example, the [path](https://pub.dev/packages/path) package. Some of these might contain Flutter-specific functionality and thus have a dependency on the Flutter framework, restricting their use to Flutter only, for example, the [fluro](https://pub.dev/packages/fluro) package.
>
> Plugin packages:
>     A specialized Dart package that contains an API written in Dart code combined with one or more platform-specific implementations.
> 
>     Plugin packages can be written for Android (using Kotlin or Java), iOS (using Swift or Objective-C), web, macOS, Windows, or Linux, or any combination thereof.

Because the new SDK is a pure dart with no platform-specific code which means the new SDK is os-agnostic. i.e. we do not request features from any platform. 

---
> And could you give a code example for these, to get an idea of what that actually means in terms of coding effort for the developer?

Sure, here is a two code snippet the first one before the transition and the second after the transition:

#### Before the transition:
``` dart 
import 'package:parse_server_sdk_flutter/parse_server_sdk.dart';

var keyParseApplicationId = '';
var keyParseServerUrl = '';
var keyParseClientKey = '';
var liveQueryUrl = '';

class InitParse {
  Future<void> init() async {
    await Parse().initialize(
      keyParseApplicationId,
      keyParseServerUrl,
      clientKey: keyParseClientKey,
      liveQueryUrl: liveQueryUrl,
      coreStore: await CoreStoreSembastImp.getInstance(),
    );
  }
}

```

#### After the transition:
``` dart 
import 'dart:async';
import 'dart:io';
import 'dart:ui' as ui;

import 'package:connectivity_plus/connectivity_plus.dart';
import 'package:flutter/foundation.dart';
import 'package:flutter/widgets.dart';
import 'package:package_info_plus/package_info_plus.dart';
import 'package:path/path.dart' as path;
import 'package:path_provider/path_provider.dart' as path_provider;

var keyParseApplicationId = '';
var keyParseServerUrl = '';
var keyParseClientKey = '';
var liveQueryUrl = '';

class InitParse with WidgetsBindingObserver {
  StreamController<void> appResumedStreamController = StreamController();

  void init() async {
    final PackageInfo packageInfo = await PackageInfo.fromPlatform();
    final appName = packageInfo.appName;
    final appVersion = packageInfo.version;
    final appPackageName = packageInfo.packageName;

    await Parse().initialize(
      keyParseApplicationId,
      keyParseServerUrl,
      clientKey: keyParseClientKey,
      liveQueryUrl: liveQueryUrl,
      appName: appName,
      appPackageName: appPackageName,
      appVersion: appVersion,
      coreStore: await CoreStoreSembastImp.getInstance(await databaseFile()),
      fileDirectory: await getTempDirectory(),
      connectivityProvider: MyConnectivityProvider(),
      locale: local(),
      appResumedStream: appResumedStreamController.stream,
    );
  }

  @override
  void didChangeAppLifecycleState(AppLifecycleState state) {
    super.didChangeAppLifecycleState(state);

    appResumedStreamController.sink.add(null);
  }

  String local() {
    return (kIsWeb ? ui.window.locale.toString() : Platform.localeName);
  }

  Future<String> databaseFile() async {
    final databaseDir = await getDatabaseDirectory();

    return path.join(databaseDir, 'parse', 'parse.db');
  }

    if (kIsWeb) {
      return '';
    }

    if (defaultTargetPlatform == TargetPlatform.iOS) {
      return (await path_provider.getLibraryDirectory()).path;
    }

    return (await path_provider.getApplicationDocumentsDirectory()).path;
  }

  Future<String> getTempDirectory() async {
    return (await path_provider.getTemporaryDirectory()).path;
  }
}

class MyConnectivityProvider extends ParseConnectivityProvider {
  @override
  Future<ParseConnectivityResult> checkConnectivity() async {
    switch (await Connectivity().checkConnectivity()) {
      case ConnectivityResult.wifi:
        return ParseConnectivityResult.wifi;
      case ConnectivityResult.mobile:
        return ParseConnectivityResult.mobile;
      case ConnectivityResult.none:
        return ParseConnectivityResult.none;
      default:
        return ParseConnectivityResult.wifi;
    }
  }

  @override
  Stream<ParseConnectivityResult> get connectivityStream {
    return Connectivity().onConnectivityChanged.map((ConnectivityResult event) {
      switch (event) {
        case ConnectivityResult.wifi:
          return ParseConnectivityResult.wifi;
        case ConnectivityResult.mobile:
          return ParseConnectivityResult.mobile;
        default:
          return ParseConnectivityResult.none;
      }
    });
  }
}


```

mtrezza:
- What is the purpose of an OS agnostic package? To me it seems more like a "core module" that is pretty useless for developers because they will always run it on _some_ OS, so they always need to wrap it into platform specific code, right?

- Why would we want to merge into a single dart package instead of a single flutter package?

- In your example code, can't we build this platform specific code into the package, so that the package contains code for multiple platforms, similar to a multiarchitecture binary? If the SDK could detect the platforms type then it can do that automatically. If not, then the developer could set it manually. For example, our API mail adapter supports multiple email services providers out of the box, the developer just needs to set the correct one and the adapter will automatically encode the payload so that the specific service provider can understand it.

Nidal-Bakir:
>The dart package seems more like a classic "core module" that is pretty useless for developers because they will always run it on some OS, so they always need to wrap it into platform specific code, right?

At the end of the day yes, will run on some OS, the idea behind this "core module" (Dart SDK) is the separation of concerns.

> Why would we want to merge into a single dart package instead of a single flutter package? What's the practical application of a pure dart package?

That was a mistake, we did not realize the amount of manual work the developers will do to set up the SDK.

> In your example code, can't we build this platform specific code into the package, so that the package contains code for multiple platforms, similar to a multiarchitecture binary? If the SDK could detect the platforms type then it can do that automatically. If not, then the developer could set it manually. For example, our API mail adapter supports multiple email services providers out of the box, the developer just needs to set the correct one and the adapter will automatically encode the payload so that the specific service provider can understand it.

yes, we can do that. by migrating to the Flutter SDK

catalunha:
Just sharing @mtrezza 
> What's the practical application of a pure dart package?

In two of my projects I open a pure Dart connection to do CRUDs and Querys in Parse Serve from b4a. Performing all the tests I need. Not in the TDD standard, but being able to access Parse Serve from b4a from a simple App in pure Dart.
So when I go to the Flutter application I already have all the CRUDs and Querys ready and it's just developing the interface.

pastordee:
> Sure, here is the full story...
> 
> > Can you help me understand why we could not provide this in the new single SDK?
> 
> Those packages depend on Flutter in their dependency to make platform-specific calls. but the new SDK should be os-agnostic. because it's pure Dart and if we use these packages in the new SKD now we are not a pure Dart package and we have a platform-specific code, even though the SDK itself not making these calls.
> 
> If we need to add these packages in the SDK we need to migrate the other way around. we need to remove the dart SDK and migrate to the Flutter SDK. then we can use packages that make calls to platform-specific functions. because in this case, a (Flutter package) package uses should use this package in the Flutter context (in a Flutter project) and not in a Dart project.
> 
> Another unexpected thing that will face the developers if we add these packages to the Dart SDK is that the developers expect the package to be pure dart so they can use it in a CLI for example or any context outside the Flutter framework, but here is the plot twist the SDK will download the entire Flutter framework for this simple console application because our dependency graph contains Flutter Framework itself as a dependency of a dependency that we are using.
> 
> The following code is from the [path_provider](https://github.com/flutter/packages/blob/11aab473cbe5819650eb43c6db9165c6e798de3c/packages/path_provider/path_provider/pubspec.yaml#L25-L27) package:
> 
> ```yaml
> dependencies:
>   flutter:
>     sdk: flutter
> ```
> 
> As you can see the path_provider package depends on the Flutter framework.
> 
> In the world of Flutter and Dart packages, we have two types:
> 
> 1. Dart package (pure Dart): the package contains only dart code no platform-specific calls, nor Flutter-related code. an example of these types of packages would be math, HTTP, Dio, or any other package that is contained in itself. i.e. can run without interaction with the host OS or the Flutter framework.
> 2. Flutter packages which in turn have tow types:
>    
>    * Flutter Package: a package that depends on the Flutter framework to do its work. you can **_use_** any package that calls functions on the host OS like path_provider.
>    * Flutter Plugin: it is a type of package that will call the underlying platform by itself like the path_provider. using platform channels for example.
> 
> [Package types](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#types)
> 
> > Packages can contain more than one kind of content:
> > Dart packages:
> > General packages written in Dart, for example, the [path](https://pub.dev/packages/path) package. Some of these might contain Flutter-specific functionality and thus have a dependency on the Flutter framework, restricting their use to Flutter only, for example, the [fluro](https://pub.dev/packages/fluro) package.
> > Plugin packages:
> > A specialized Dart package that contains an API written in Dart code combined with one or more platform-specific implementations.
> > ```
> > Plugin packages can be written for Android (using Kotlin or Java), iOS (using Swift or Objective-C), web, macOS, Windows, or Linux, or any combination thereof.
> > ```
> 
> Because the new SDK is a pure dart with no platform-specific code which means the new SDK is os-agnostic. i.e. we do not request features from any platform.
> 
> > And could you give a code example for these, to get an idea of what that actually means in terms of coding effort for the developer?
> 
> Sure, here is a two code snippet the first one before the transition and the second after the transition:
> 
> #### Before the transition:
> ```dart
> import 'package:parse_server_sdk_flutter/parse_server_sdk.dart';
> 
> var keyParseApplicationId = '';
> var keyParseServerUrl = '';
> var keyParseClientKey = '';
> var liveQueryUrl = '';
> 
> class InitParse {
>   Future<void> init() async {
>     await Parse().initialize(
>       keyParseApplicationId,
>       keyParseServerUrl,
>       clientKey: keyParseClientKey,
>       liveQueryUrl: liveQueryUrl,
>       coreStore: await CoreStoreSembastImp.getInstance(),
>     );
>   }
> }
> ```
> 
> #### After the transition:
> ```dart
> import 'dart:async';
> import 'dart:io';
> import 'dart:ui' as ui;
> 
> import 'package:connectivity_plus/connectivity_plus.dart';
> import 'package:flutter/foundation.dart';
> import 'package:flutter/widgets.dart';
> import 'package:package_info_plus/package_info_plus.dart';
> import 'package:parse_server_sdk/parse_server_sdk.dart';
> import 'package:path/path.dart' as path;
> import 'package:path_provider/path_provider.dart' as path_provider;
> 
> var keyParseApplicationId = '';
> var keyParseServerUrl = '';
> var keyParseClientKey = '';
> var liveQueryUrl = '';
> 
> class InitParse with WidgetsBindingObserver {
>   StreamController<void> appResumedStreamController = StreamController();
> 
>   void init() async {
>     final PackageInfo packageInfo = await PackageInfo.fromPlatform();
>     final appName = packageInfo.appName;
>     final appVersion = packageInfo.version;
>     final appPackageName = packageInfo.packageName;
> 
>     await Parse().initialize(
>       keyParseApplicationId,
>       keyParseServerUrl,
>       clientKey: keyParseClientKey,
>       liveQueryUrl: liveQueryUrl,
>       appName: appName,
>       appPackageName: appPackageName,
>       appVersion: appVersion,
>       coreStore: await CoreStoreSembastImp.getInstance(await databaseFile()),
>       fileDirectory: await getTempDirectory(),
>       connectivityProvider: MyConnectivityProvider(),
>       locale: local(),
>       appResumedStream: appResumedStreamController.stream,
>     );
>   }
> 
>   @override
>   void didChangeAppLifecycleState(AppLifecycleState state) {
>     super.didChangeAppLifecycleState(state);
> 
>     appResumedStreamController.sink.add(null);
>   }
> 
>   String local() {
>     return (kIsWeb ? ui.window.locale.toString() : Platform.localeName);
>   }
> 
>   Future<String> databaseFile() async {
>     final databaseDir = await getDatabaseDirectory();
> 
>     return path.join(databaseDir, 'parse', 'parse.db');
>   }
> 
>   Future<String> getDatabaseDirectory() async {
>     if (kIsWeb) {
>       return '';
>     }
> 
>     if (defaultTargetPlatform == TargetPlatform.iOS) {
>       return (await path_provider.getLibraryDirectory()).path;
>     }
> 
>     return (await path_provider.getApplicationDocumentsDirectory()).path;
>   }
> 
>   Future<String> getTempDirectory() async {
>     return (await path_provider.getTemporaryDirectory()).path;
>   }
> }
> 
> class MyConnectivityProvider extends ParseConnectivityProvider {
>   @override
>   Future<ParseConnectivityResult> checkConnectivity() async {
>     switch (await Connectivity().checkConnectivity()) {
>       case ConnectivityResult.wifi:
>         return ParseConnectivityResult.wifi;
>       case ConnectivityResult.mobile:
>         return ParseConnectivityResult.mobile;
>       case ConnectivityResult.none:
>         return ParseConnectivityResult.none;
>       default:
>         return ParseConnectivityResult.wifi;
>     }
>   }
> 
>   @override
>   Stream<ParseConnectivityResult> get connectivityStream {
>     return Connectivity().onConnectivityChanged.map((ConnectivityResult event) {
>       switch (event) {
>         case ConnectivityResult.wifi:
>           return ParseConnectivityResult.wifi;
>         case ConnectivityResult.mobile:
>           return ParseConnectivityResult.mobile;
>         default:
>           return ParseConnectivityResult.none;
>       }
>     });
>   }
> }
> ```
> 
> ## How to avoid this?
> We have two options:
> 
> * If we still insist on one package/repo, then we need to migrate to the Flutter SDK and embed the Dart SDK in the Flutter SDK. But then the SDK can not run without the Flutter framework.
> * keep everything as is. two packages.

Hi guys, as I commented before this was the exact reason the package was split into two in the first please again I’ve been using it from day one and in the beginning it was together but because of platform-specific that’s why it was divided. I hope this helps.

pastordee:
> Sure, here is the full story...
> 
> > Can you help me understand why we could not provide this in the new single SDK?
> 
> Those packages depend on Flutter in their dependency to make platform-specific calls. but the new SDK should be os-agnostic. because it's pure Dart and if we use these packages in the new SKD now we are not a pure Dart package and we have a platform-specific code, even though the SDK itself not making these calls.
> 
> If we need to add these packages in the SDK we need to migrate the other way around. we need to remove the dart SDK and migrate to the Flutter SDK. then we can use packages that make calls to platform-specific functions. because in this case, a (Flutter package) package uses should use this package in the Flutter context (in a Flutter project) and not in a Dart project.
> 
> Another unexpected thing that will face the developers if we add these packages to the Dart SDK is that the developers expect the package to be pure dart so they can use it in a CLI for example or any context outside the Flutter framework, but here is the plot twist the SDK will download the entire Flutter framework for this simple console application because our dependency graph contains Flutter Framework itself as a dependency of a dependency that we are using.
> 
> The following code is from the [path_provider](https://github.com/flutter/packages/blob/11aab473cbe5819650eb43c6db9165c6e798de3c/packages/path_provider/path_provider/pubspec.yaml#L25-L27) package:
> 
> ```yaml
> dependencies:
>   flutter:
>     sdk: flutter
> ```
> 
> As you can see the path_provider package depends on the Flutter framework.
> 
> In the world of Flutter and Dart packages, we have two types:
> 
> 1. Dart package (pure Dart): the package contains only dart code no platform-specific calls, nor Flutter-related code. an example of these types of packages would be math, HTTP, Dio, or any other package that is contained in itself. i.e. can run without interaction with the host OS or the Flutter framework.
> 2. Flutter packages which in turn have tow types:
>    
>    * Flutter Package: a package that depends on the Flutter framework to do its work. you can **_use_** any package that calls functions on the host OS like path_provider.
>    * Flutter Plugin: it is a type of package that will call the underlying platform by itself like the path_provider. using platform channels for example.
> 
> [Package types](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#types)
> 
> > Packages can contain more than one kind of content:
> > Dart packages:
> > General packages written in Dart, for example, the [path](https://pub.dev/packages/path) package. Some of these might contain Flutter-specific functionality and thus have a dependency on the Flutter framework, restricting their use to Flutter only, for example, the [fluro](https://pub.dev/packages/fluro) package.
> > Plugin packages:
> > A specialized Dart package that contains an API written in Dart code combined with one or more platform-specific implementations.
> > ```
> > Plugin packages can be written for Android (using Kotlin or Java), iOS (using Swift or Objective-C), web, macOS, Windows, or Linux, or any combination thereof.
> > ```
> 
> Because the new SDK is a pure dart with no platform-specific code which means the new SDK is os-agnostic. i.e. we do not request features from any platform.
> 
> > And could you give a code example for these, to get an idea of what that actually means in terms of coding effort for the developer?
> 
> Sure, here is a two code snippet the first one before the transition and the second after the transition:
> 
> #### Before the transition:
> ```dart
> import 'package:parse_server_sdk_flutter/parse_server_sdk.dart';
> 
> var keyParseApplicationId = '';
> var keyParseServerUrl = '';
> var keyParseClientKey = '';
> var liveQueryUrl = '';
> 
> class InitParse {
>   Future<void> init() async {
>     await Parse().initialize(
>       keyParseApplicationId,
>       keyParseServerUrl,
>       clientKey: keyParseClientKey,
>       liveQueryUrl: liveQueryUrl,
>       coreStore: await CoreStoreSembastImp.getInstance(),
>     );
>   }
> }
> ```
> 
> #### After the transition:
> ```dart
> import 'dart:async';
> import 'dart:io';
> import 'dart:ui' as ui;
> 
> import 'package:connectivity_plus/connectivity_plus.dart';
> import 'package:flutter/foundation.dart';
> import 'package:flutter/widgets.dart';
> import 'package:package_info_plus/package_info_plus.dart';
> import 'package:parse_server_sdk/parse_server_sdk.dart';
> import 'package:path/path.dart' as path;
> import 'package:path_provider/path_provider.dart' as path_provider;
> 
> var keyParseApplicationId = '';
> var keyParseServerUrl = '';
> var keyParseClientKey = '';
> var liveQueryUrl = '';
> 
> class InitParse with WidgetsBindingObserver {
>   StreamController<void> appResumedStreamController = StreamController();
> 
>   void init() async {
>     final PackageInfo packageInfo = await PackageInfo.fromPlatform();
>     final appName = packageInfo.appName;
>     final appVersion = packageInfo.version;
>     final appPackageName = packageInfo.packageName;
> 
>     await Parse().initialize(
>       keyParseApplicationId,
>       keyParseServerUrl,
>       clientKey: keyParseClientKey,
>       liveQueryUrl: liveQueryUrl,
>       appName: appName,
>       appPackageName: appPackageName,
>       appVersion: appVersion,
>       coreStore: await CoreStoreSembastImp.getInstance(await databaseFile()),
>       fileDirectory: await getTempDirectory(),
>       connectivityProvider: MyConnectivityProvider(),
>       locale: local(),
>       appResumedStream: appResumedStreamController.stream,
>     );
>   }
> 
>   @override
>   void didChangeAppLifecycleState(AppLifecycleState state) {
>     super.didChangeAppLifecycleState(state);
> 
>     appResumedStreamController.sink.add(null);
>   }
> 
>   String local() {
>     return (kIsWeb ? ui.window.locale.toString() : Platform.localeName);
>   }
> 
>   Future<String> databaseFile() async {
>     final databaseDir = await getDatabaseDirectory();
> 
>     return path.join(databaseDir, 'parse', 'parse.db');
>   }
> 
>   Future<String> getDatabaseDirectory() async {
>     if (kIsWeb) {
>       return '';
>     }
> 
>     if (defaultTargetPlatform == TargetPlatform.iOS) {
>       return (await path_provider.getLibraryDirectory()).path;
>     }
> 
>     return (await path_provider.getApplicationDocumentsDirectory()).path;
>   }
> 
>   Future<String> getTempDirectory() async {
>     return (await path_provider.getTemporaryDirectory()).path;
>   }
> }
> 
> class MyConnectivityProvider extends ParseConnectivityProvider {
>   @override
>   Future<ParseConnectivityResult> checkConnectivity() async {
>     switch (await Connectivity().checkConnectivity()) {
>       case ConnectivityResult.wifi:
>         return ParseConnectivityResult.wifi;
>       case ConnectivityResult.mobile:
>         return ParseConnectivityResult.mobile;
>       case ConnectivityResult.none:
>         return ParseConnectivityResult.none;
>       default:
>         return ParseConnectivityResult.wifi;
>     }
>   }
> 
>   @override
>   Stream<ParseConnectivityResult> get connectivityStream {
>     return Connectivity().onConnectivityChanged.map((ConnectivityResult event) {
>       switch (event) {
>         case ConnectivityResult.wifi:
>           return ParseConnectivityResult.wifi;
>         case ConnectivityResult.mobile:
>           return ParseConnectivityResult.mobile;
>         default:
>           return ParseConnectivityResult.none;
>       }
>     });
>   }
> }
> ```
> 
> ## How to avoid this?
> We have two options:
> 
> * If we still insist on one package/repo, then we need to migrate to the Flutter SDK and embed the Dart SDK in the Flutter SDK. But then the SDK can not run without the Flutter framework.
> * keep everything as is. two packages.

Sorry guys, I think we should keep the two packages. My reason for that is the dart version is the core and the flatter is built on top of that and anyone can build another package on top of it say a sign up package on top of the dart and people will have the choice to use it or not. Again anyone using the dart version do not need  the flatter package , but if you use the flutter package, you will still need to import the data version into it. So I think we should leave the two packages and just allow people to build packages on top of the Dart package That’s my view.

Nidal-Bakir:
@mtrezza 
We have wasted a lot of time on this. We need to make up our minds. There are a lot of open issues and feature requests besides improving the SDK, adding some documentation in the code, adding any missing features, and so on ...

Currently, the option is:

1. drop the Flutter SDK and keep the Dart SDK (this PR)
2. mege the Dart SDK in Fluttter SDK
3. keep everything as it is. (the best option)
4. defer this to another time

mbfakourii:
In my opinion, we should solve the problem with some explanations, I think that the problem will be solved by explaining the data and the developers can adapt.

mbfakourii:
We spent a lot of time on this PR, I don't think it's right to skip it

pastordee:
> We spent a lot of time on this PR, I don't think it's right to skip it

I don’t think it’s been skip is just that we need to take time and really think about it but in the meantime we should continue to update the flutter package .

mtrezza:
Thanks everyone for their input in this discussion. The time is well spent if we come to the right conclusion.

My previous point is still unclear to me:

> Why would we want to merge into a single dart package instead of a single flutter package? What's the practical application of a pure dart package?

And what's the practical downside of offering a flutter package instead of a dart package?

mtrezza:
Got it, thanks. And why can't we include the platform specific code that is mentioned in "after transition" in the [comment](https://github.com/parse-community/Parse-SDK-Flutter/pull/852#issuecomment-1473357297) in the dart package to make it a "universal package"?

Nidal-Bakir:
> And can't we include the platform specific code that is mentioned in https://github.com/parse-community/Parse-SDK-Flutter/pull/852#issuecomment-1473357297 in the dart package to make it a "universal package"?

Because any platform specific code requires the Flutter framework to be present. any  platform specific call is done through the Flutter framework 


> The code contains flutter imports - could they be imported only optionally somehow?

These imports are declared outside the SDK. i.e. the developer is writing this. 


the developer will write this.
 
> #### After the transition:
> 
> ```dart
> import 'dart:async';
> import 'dart:io';
> import 'dart:ui' as ui;
> 
> import 'package:connectivity_plus/connectivity_plus.dart';
> import 'package:flutter/foundation.dart';
> import 'package:flutter/widgets.dart';
> import 'package:package_info_plus/package_info_plus.dart';
> import 'package:parse_server_sdk/parse_server_sdk.dart';
> import 'package:path/path.dart' as path;
> import 'package:path_provider/path_provider.dart' as path_provider;
> 
> var keyParseApplicationId = '';
> var keyParseServerUrl = '';
> var keyParseClientKey = '';
> var liveQueryUrl = '';
> 
> class InitParse with WidgetsBindingObserver {
>   StreamController<void> appResumedStreamController = StreamController();
> 
>   void init() async {
>     final PackageInfo packageInfo = await PackageInfo.fromPlatform();
>     final appName = packageInfo.appName;
>     final appVersion = packageInfo.version;
>     final appPackageName = packageInfo.packageName;
> 
>     await Parse().initialize(
>       keyParseApplicationId,
>       keyParseServerUrl,
>       clientKey: keyParseClientKey,
>       liveQueryUrl: liveQueryUrl,
>       appName: appName,
>       appPackageName: appPackageName,
>       appVersion: appVersion,
>       coreStore: await CoreStoreSembastImp.getInstance(await databaseFile()),
>       fileDirectory: await getTempDirectory(),
>       connectivityProvider: MyConnectivityProvider(),
>       locale: local(),
>       appResumedStream: appResumedStreamController.stream,
>     );
>   }
> 
>   @override
>   void didChangeAppLifecycleState(AppLifecycleState state) {
>     super.didChangeAppLifecycleState(state);
> 
>     appResumedStreamController.sink.add(null);
>   }
> 
>   String local() {
>     return (kIsWeb ? ui.window.locale.toString() : Platform.localeName);
>   }
> 
>   Future<String> databaseFile() async {
>     final databaseDir = await getDatabaseDirectory();
> 
>     return path.join(databaseDir, 'parse', 'parse.db');
>   }
> 
>   Future<String> getDatabaseDirectory() async {
>     if (kIsWeb) {
>       return '';
>     }
> 
>     if (defaultTargetPlatform == TargetPlatform.iOS) {
>       return (await path_provider.getLibraryDirectory()).path;
>     }
> 
>     return (await path_provider.getApplicationDocumentsDirectory()).path;
>   }
> 
>   Future<String> getTempDirectory() async {
>     return (await path_provider.getTemporaryDirectory()).path;
>   }
> }
> 
> class MyConnectivityProvider extends ParseConnectivityProvider {
>   @override
>   Future<ParseConnectivityResult> checkConnectivity() async {
>     switch (await Connectivity().checkConnectivity()) {
>       case ConnectivityResult.wifi:
>         return ParseConnectivityResult.wifi;
>       case ConnectivityResult.mobile:
>         return ParseConnectivityResult.mobile;
>       case ConnectivityResult.none:
>         return ParseConnectivityResult.none;
>       default:
>         return ParseConnectivityResult.wifi;
>     }
>   }
> 
>   @override
>   Stream<ParseConnectivityResult> get connectivityStream {
>     return Connectivity().onConnectivityChanged.map((ConnectivityResult event) {
>       switch (event) {
>         case ConnectivityResult.wifi:
>           return ParseConnectivityResult.wifi;
>         case ConnectivityResult.mobile:
>           return ParseConnectivityResult.mobile;
>         default:
>           return ParseConnectivityResult.none;
>       }
>     });
>   }
> }
> ```

Nidal-Bakir:
Another convention in the Flutter/Dart packages world is that if any package on the pub.dev has the "Dart" in its name that means it does not need flutter to run.

And the same thing goes for any package that has the "Flutter" name in it. its like a hint that the package needs flutter to run

That's why we can not convert a Dart package to Flutter one. (We can but we should not)

mtrezza:
Well, I think we can put this PR on hold for now as further deliberation seems to be needed.

mbfakourii:
@Nidal-Bakir 

I implemented a new method that the flutter tasks can only use the following code. what is your opinion ?

```dart
import 'package:parse_server_sdk/src/utils/parse_server_sdk_flutter.dart';
import 'dart:ui' as ui;

main() async {
  Parse().initializeFlutter(
      "appId",
      "serverUrl",
      packageInfo: await PackageInfo.fromPlatform(),
      ui: ui,
      connectivity: await Connectivity(),
      pathProvider: pathProvider,
      coreStore:  await CoreStoreSharedPrefsImp.getInstance(
          sharedPreferences: (await SharedPreferences.getInstance())),
  );
}

```

mtrezza:
How does this address the concerns that have been voiced?

mbfakourii:
> How does this address the concerns that have been voiced?

In this method, all the initial values that were in the Flutter version are set
With this, we actually support the Flutter version in Dart and also use pure Dart codes

mtrezza:
Ok, I think the biggest concern with this PR is the additional code that developers would need to write to adapt a single dart package to their OS specific APIs. What could we do about that?

mbfakourii:
> Ok, I think the biggest concern with this PR is the additional code that developers would need to write to adapt a single dart package to their OS specific APIs. What could we do about that?

I said above that the best way is to explain line by line, I will create a separate document for this topic

mtrezza:
Could we go back to what the actual pain point is that led to merging the 2 packages?

mbfakourii:
> Could we go back to what the actual pain point is that led to the idea of merging the 2 packages? Is that pain on the developers' side or on the maintainers' side, or both?

More side maintainers
But developers choose a library with more confidence

mtrezza:
Could you outline what the current difficulties for maintainers are? Apologies if that has been posted already somewhere.

mbfakourii:
> Could you outline what the current difficulties for maintainers are? Apologies if that has been posted already somewhere.
> 
> Also, could we look at how others are solving this? If we look at other SDK providers, what do they offer, 2 packages or only dart or only flutter?
> 
> Looking at firebase as an example, they seem to offer:
> 
> * [dart core package](https://pub.dev/packages/firebase_core)
> * [dart OS-specific packages](https://pub.dev/packages?q=dependency%3Afirebase_core)
> * flutter package

I got it . I think it is better to close this PR

mtrezza:
Are you sure? You seem to have put a lot of work into the PR. I'm not suggesting either way, I'm just trying to facilitate the conversation.

mbfakourii:
> Are you sure? You seem to have put a lot of work into the PR. I'm not suggesting either way, I'm just trying to facilitate the conversation to hopefully reach more of a consensus.

I am not sure. I spent about a day for this PR, but what to do when friends don't support

mtrezza:
Alright, so the conclusion is that the status quo is fine, or can we learn anything from the other packages?

mbfakourii:
> Alright, so the conclusion is that the status quo is fine, or can we learn anything from the other packages?

The current situation is good

mtrezza:
Got it, well thanks for putting in all the work into the PRs and all others for their input into this discussion.

Issue #843, opened 2023-03-04
Incrementing or Decrementing existing value for the first time will remove it


Nidal-Bakir:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
using `set` to set a value and then trying to increment or decrement that value will result in removing the value and the amount parameter from the increment or decrement functions will replace it after this point any call to increment or decrement functions will work fine.

### Steps to reproduce
1. create `ParseObject`
2. set a value `set("key",2)`
3. increment that value using the`setIncrement("key",5)` function
4. get the value using the value key with `get("key")`
5. the result of calling `get("key")` is `5` which is not expected 
6. noteis that 5 is the amount is passed to setIncrement function

code snippet:
```dart 
final dietPlansObject = ParseObject("Diet_Plans");

dietPlansObject.set("key", 2);

final anyAmount = 5;
dietPlansObject.setIncrement("key", anyAmount);
// dietPlansObject.setDecrement("key", anyAmount);

final value = dietPlansObject.get("key");

print(value); // 5 but it should be 7
``` 

### Actual Outcome
Not considering the already existing value in the first call to `setIncrement` or `setDecrement`

### Expected Outcome
check for already existing value and use it.

### Environment

Parse Flutter SDK
- SDK version: `3.1.15`
- Operating system version: `any`

Server
- Parse Server version: `any`

Nidal-Bakir:
Looks like the SDK is not using the [Escape Codes](https://docs.microfocus.com/OMi/10.62/Content/OMi/ExtGuide/ExtApps/URL_encoding.htm) when encoding the request.
I will investigate more in this.


```dart
// this will work 

  final url = Uri.parse(
                  'https://parseapi.back4app.com/classes/_User?where={"email": "gg%2Br@g.com"}');
              final res = await http.get(url, headers: {
                'X-Parse-Application-Id': '',
[...]

Issue #842, opened 2023-03-04
Incrementing or Decrementing a value and then calling `get` returns _Map<String, dynamic>


Nidal-Bakir:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
Incrementing or Decrementing a value using `setDecrement(key,amount)` or `setIncrement(key,amount)` and then calling `get(key)` returns `_Map<String, dynamic> `

related to: #834, #696

### Steps to reproduce

1. create `ParseObject`
2. increment or decrement a value using `setIncrement` or `setDecrement` functions
3. get the value using the value key with `get("key")`
4. the result of calling `get("key")` is `Map` which is not expected

code snippet:
```dart 
final dietPlansObject = ParseObject("Diet_Plans");

final anyAmount = 1;
dietPlansObject.setIncrement("key", anyAmount);
// dietPlansObject.setDecrement("key", anyAmount);

final value = dietPlansObject.get("key");

print(value.runtimeType); // _Map<String, dynamic>
``` 

### Actual Outcome
`_Map<String, dynamic>`

### Expected Outcome
`num`

### Environment

Parse Flutter SDK
- SDK version: `3.1.15`
- Operating system version: `any`

Server
- Parse Server version: `any`

### Logs
From tests error:

> Expected: <Instance of 'num'>
>   Actual: {'__op': 'Increment', 'amount': 1.0}
>    Which: is not an instance of 'num'
> package:test_api                                 expect
> test/src/objects/parse_object_test.dart 1180:11  main.<fn>.<fn>.<fn>

mtrezza:
@Nidal-Bakir I'm removing you from all issues as assignee. We do not use the assignment feature.

Nidal-Bakir:
Fix via #860

Issue #834, opened 2023-03-01
adding or removing values from or into an array and then calling `get` returns  `_Map<String, dynamic>`


Nidal-Bakir:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
Adding or removing values from or into an array using (`setAdd`, `setAddAll`, `setAddUnique`, `setAddAllUnique`, `setRemove`, `setRemoveAll`) in `ParseObject` and then calling `get` returns  `_Map<String, dynamic>`

related to: #696, #842 

### Steps to reproduce

1. create `ParseObject`
2. add a value to an array using `setAdd` function for example 
3. get the array using the array key with `get("key")`
4. the result of calling `get("key")` is `Map` which is not expected 

code snippet:
```dart 
final dietPlansObject = ParseObject("Diet_Plans");

dietPlansObject.setAdd("arrayKey", 1);

final array = dietPlansObject.get("arrayKey");

print(array.runtimeType); // _Map<String, dynamic>
``` 

### Actual Outcome
`_Map<String, dynamic>`

### Expected Outcome
An `Iterable` that contains all the added values and does not contain the removed ones

### Environment

Parse Flutter SDK
- SDK version: `3.1.15`
- Operating system version: `any`

Server
- Parse Server version: `any`

### Logs

> type '_Map<String, dynamic>' is not a subtype of type 'List<dynamic>?' in type cast
package:parse_server_sdk/src/objects/parse_base.dart 223:36  ParseBase.get
test/src/objects/parse_object_test.dart 711:39               main.<fn>.<fn>.<fn>
parse_object_test.dart:711

Nidal-Bakir:
I just find it out while writing tests for `ParseObject` class.

I have written all the tests to make sure that will not happen again we just need the fix.

mtrezza:
[...]

Issue #696, opened 2021-12-14
ParseRelation#query - Unhandled Exception: type 'ParseObject' is not a subtype of type *


kutear:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
When i get a ParseRelation(`relation`) from a subClass of ParseObject, then build a QueryBuilder(`queryBuilder`) by `relation.getQuery()`, and invoke `builder.query<Wallet>()` to get a response.  at this time i get some error.
```
E/flutter ( 7948): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: type 'ParseObject' is not a subtype of type 'Wallet'
E/flutter ( 7948): #0      _ParseResponseBuilder._handleSuccess (package:parse_server_sdk/src/objects/response/parse_response_builder.dart:93:59)
E/flutter ( 7948): #1      _ParseResponseBuilder.handleResponse (package:parse_server_sdk/src/objects/response/parse_response_builder.dart:29:14)
E/flutter ( 7948): #2      handleResponse (package:parse_server_sdk/src/objects/response/parse_response_utils.dart:8:31)
E/flutter ( 7948): #3      ParseObject.query (package:parse_server_sdk/src/objects/parse_object.dart:419:14)
```

### Steps to reproduce
```
    class Plan extends ParseObject {
    ...
    }

    final plan = ...;
    ParseRelation<Wallet> relation = plan.getRelation("wallets");
    QueryBuilder builder = relation.getQuery();
    final response = await builder.query<Wallet>();
    if (response.success) {
      // do something
    }



```

### Actual Outcome
```
Unhandled Exception: type 'ParseObject' is not a subtype of type 'Wallet'
```

### Expected Outcome
query success and get  a list of Wallet

### Environment
```
  parse_server_sdk_flutter: 3.1.0
```

Parse Flutter SDK
- SDK version: `3.1.0`
[...]

PR #860, opened 2023-03-19
fix: Setting atomic operation on Parse Object returns operation instead of prospective value

Nidal-Bakir:

### Issue Description

Closes: #843, #842, #834, #696

### Approach

Create a new Class/DataType to handle any parse operation

 * setAdd
 * setAddUnique
 * setAddAll
* setAddAllUnique
* setRemove
 * setRemoveAll
* setIncrement
 * setDecrement
 * addRelation
 * removeRelation

Why does this PR fixes all these issues? 
Well, they all suffer from the same bug (using Maps to store operations). Because every operation needs to be in a new object and the old way of using maps should be removed, keeping both and fixing them one by one could create unexpected behaviors because we would need to use both of them while fixing the bugs.
 

### TODOs before merging

- [x] Add tests
- [x] A changelog entry


Nidal-Bakir:
I won't be available on GitHub for a few days as I am building a test app for a company I could potentially work with. They have given me some requirements that I need to fulfill, and my deadline is next week I will come back to this PR once I'm done with the test app.

Nidal-Bakir:
### set Add/AddAll functions:

 NOTE: notice how the old and the new API have the same result when converting the object to JSON for saving it

#### example 1.1:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setAdd('arr', 1);
    dietPlansObject.setAdd('arr', 2);

    print(dietPlansObject.get('arr')); // {__op: Add, objects: [1, 2]}

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [1, 2]}}

* new API output:

    ``` dart
      final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setAdd('arr', 1);
    dietPlansObject.setAdd('arr', 2);

    print(dietPlansObject.get('arr')); // [1, 2]

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [1, 2]}}
    ```

notice how the new API will output the correct type (List) with the added data.

#### example 1.2:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.set('arr', [1, 2]);

    dietPlansObject.setAddAll('arr', [3, 4]);

    dietPlansObject.setRemove('arr', 2);

    print(dietPlansObject.get('arr')); // [1, 2, 3, 4]

    print(dietPlansObject.toJson(forApiRQ: true));  // {arr: [1, 2, 3, 4]}
   ```
* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.set('arr', [1, 2]);

    dietPlansObject.setAddAll('arr', [3, 4]);

    dietPlansObject.setRemove('arr', 2); // will throw (Remove operation is invalid after Add operation)

    print(dietPlansObject.get('arr'));  

    print(dietPlansObject.toJson(forApiRQ: true));
    ```

notice how the new API will throw an exception when two conflicting operations are performed on the same key. while the old API did nothing about it and did not remove nor throw an exception

#### example 1.3:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setAdd('arr', 3);

    print(dietPlansObject.get('arr')); // {__op: Add, objects: [3]} 

    print(dietPlansObject.toJson(forApiRQ: true)); //  {arr: {__op: Add, objects: [3]}}
    ```

* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setAdd('arr', 3);

    print(dietPlansObject.get('arr')); // [1, 2, 3] 

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [3]}}
    ```

The new API will account for the List coming from the server and build the new data based on it. on the other hand, the old API will return a Map and it is the wrong data type. and all the stored values in the server are gone in this object and the user can not get them. so there is no way to know what the List will be after the add operation! the user needs to save the object and then refetch it from the server to see the result list!

Nidal-Bakir:
### set Add/AddAll functions:

 NOTE: notice how the old and the new API have the same result when converting the object to JSON for saving it

#### example 1.1:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setAdd('arr', 1);
    dietPlansObject.setAdd('arr', 2);

    print(dietPlansObject.get('arr')); // {__op: Add, objects: [1, 2]}

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [1, 2]}}

* new API output:

    ``` dart
      final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setAdd('arr', 1);
    dietPlansObject.setAdd('arr', 2);

    print(dietPlansObject.get('arr')); // [1, 2]

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [1, 2]}}
    ```

notice how the new API will output the correct type (List) with the added data.

#### example 1.2:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.set('arr', [1, 2]);

    dietPlansObject.setAddAll('arr', [3, 4]);

    dietPlansObject.setRemove('arr', 2);

    print(dietPlansObject.get('arr')); // [1, 2, 3, 4]

    print(dietPlansObject.toJson(forApiRQ: true));  // {arr: [1, 2, 3, 4]}
   ```
* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.set('arr', [1, 2]);

    dietPlansObject.setAddAll('arr', [3, 4]);

    dietPlansObject.setRemove('arr', 2); // will throw (Remove operation is invalid after Add operation)

    print(dietPlansObject.get('arr'));  

    print(dietPlansObject.toJson(forApiRQ: true));
    ```

notice how the new API will throw an exception when two conflicting operations are performed on the same key. while the old API did nothing about it and did not remove nor throw an exception

#### example 1.3:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setAdd('arr', 3);

    print(dietPlansObject.get('arr')); // {__op: Add, objects: [3]} 

    print(dietPlansObject.toJson(forApiRQ: true)); //  {arr: {__op: Add, objects: [3]}}
    ```

* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setAdd('arr', 3);

    print(dietPlansObject.get('arr')); // [1, 2, 3] 

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Add, objects: [3]}}
    ```

The new API will account for the List coming from the server and build the new data based on it. on the other hand, the old API will return a Map and it is the wrong data type. and all the stored values in the server are gone in this object and the user can not get them. so there is no way to know what the List will be after the add operation! the user needs to save the object and then refetch it from the server to see the result list!

mbfakourii (approved):
I didn't see any problem, but there are many changes, @Nidal-Bakir don't you think it's better to give a beta version?

Nidal-Bakir:
### set Remove/RemoveAll functions:

 NOTE: notice how the old and the new API have the same result when converting the object to JSON for saving it

#### example 2.1:

 * old API output:
    ``` dart 
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2, 3, 3]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setRemove('arr', 3);

    print(dietPlansObject.get('arr')); // {__op: Remove, objects: [3]}

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Remove, objects: [3]}}
    ```

* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");
    dietPlansObject.fromJson({'arr': [1, 2, 3, 3]}); // assume the "arr" key is saved and coming from the server

    dietPlansObject.setRemove('arr', 3);

    print(dietPlansObject.get('arr')); // [1, 2]

    print(dietPlansObject.toJson(forApiRQ: true)); // {arr: {__op: Remove, objects: [3]}}
    ```

The new API will account for the List coming from the server and build the new data based on it. on the other hand, the old API will return a Map, which is the wrong data type. and all the stored values in the server are gone in this object and the user can not get them. so there is no way to know what the List will be after the Remove operation! The user must save the object and re-fetch it from the server to see the result list!

Nidal-Bakir:
## Number operations (Increment/Decrement)

**setMode**: is the intention to set the number as-is.

In set mode, the intention is to set the number as-is, without performing any operations on it. For example, calling set with a number as an argument like this: `dietPlansObject.set('myNum',1);`. This will create a `_ParseNumber` with `setMode`, meaning that the user can perform any (number operation: i.e.  Increment/Decrement) on this number without any conflicts. 

### set Increment/Decrement functions:

#### example 1:

 * old API output:
 

    ``` dart
    1 final dietPlansObject = ParseObject("Diet_Plans");

    2 dietPlansObject.set('numKey', 1);

    3 dietPlansObject.setIncrement('numKey', 3);
    4 dietPlansObject.setDecrement('numKey', 1);

    5 print(dietPlansObject.get('numKey')); // {__op: Increment, amount: 3}

    6 print(dietPlansObject.toJson(forApiRQ: true)); // {numKey: {__op: Increment, amount: 3}}

    ```


* new API output:

    ``` dart
  1  final dietPlansObject = ParseObject("Diet_Plans");

  2  dietPlansObject.set('numKey', 1);

  3  dietPlansObject.setIncrement('numKey', 3);
  4  dietPlansObject.setDecrement('numKey', 1);

  5  print(dietPlansObject.get('numKey')); // 3

  6  print(dietPlansObject.toJson(forApiRQ: true)); // {numKey: 3}

    ```

The new API will recognize that the user will set a new value for the "numKey" key and not perform an operation on it. so the setIncrement/setDecrement will work locally and manipulate the setted value at line 2. if the user need to reform an operation over the number like `Increment` they will not set it first using `set` but they will call `setIncrement` directly without the ` set` at line 2 in this case the user needs to perform an operation on the number `Increment` we can see that in the next example...

#### example 2:

 * old API output:
 

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setIncrement('numKey', 3);

    print(dietPlansObject.get('numKey')); // {__op: Increment, amount: 3}

    print(dietPlansObject.toJson(forApiRQ: true)); // {numKey: {__op: Increment, amount: 3}}

    ```


* new API output:

    ``` dart
    final dietPlansObject = ParseObject("Diet_Plans");

    dietPlansObject.setIncrement('numKey', 3);

    print(dietPlansObject.get('numKey')); // 3

    print(dietPlansObject.toJson(forApiRQ: true)); // {numKey: {__op: Increment, amount: 3.0}}

    ```
Here the user did not set a value for the "numKey" key but performed an operation directly in this case the new API will send an operation request `Increment` and the same goes if the value comes from the server.  When the comes from the server the new API will always send operation requests and will **never** enter the `setMode`.

Nidal-Bakir:
## ParseRelation 

....

Nidal-Bakir:
## ParseSaveStateAwareChild

....

Nidal-Bakir:

I think I have explained every aspect of this PR and all the fixes in it. We can discuss anything about it. And if there is anything I should explain more, please tell me.

I will start adding tests and documentation for the code.

Nidal-Bakir:

We are ready to go! everything is in its place tested and documented

I need to add a guide to migrate from the previous version and explain what changed and what has been fixed.

Where should I put such a guide?

mtrezza:
Amazing the amount of work you put into this! It seems this will cause the release of `dart-5.0.0`. If the changes are minor can be explained in the property and method annotations, when I'd suggest to only add it there, to make it available directly in the IDE without the developer having to open / look for another document. If you think a migration guide makes sense, then that would go under `./packages/dart/MIGRATION-5.0.0.md`.

Nidal-Bakir:
Well, to be honest, this bug fix in this pull request is not a breaking change at all. It simply corrects bugs and incorrect behaviors in the SDK. My concern is that these bugs and behaviors have existed in the SDK for quite some time, and developers have adapted their code to work around these issues. By fixing these bugs and behaviors, we may unintentionally break their existing code instead of fixing it. This is why I labeled this PR as 'breaking'.  So while not semantically a breaking change, the practical effects could be quite breaking for some. I would call it a _Bug fix release with potential breaking_

I can provide an explanation of each fix, including examples of previous issues and the expected corrected behavior, in the change log as a summary of all the fixes. Alternatively, I can create a `MIGRATION-5.0.0.md` file, in which I explain each fix, what the old API was doing wrong, and what the new API will return.

> If the changes are minor can be explained in the property and method annotation

we can not do that, because we can not say in the  method documentation something like this:

```
/// If the return type is an Array, previously will return a Map instead of
/// the requested array if an operation was performed on it, but now will return a List 
object.get('someKey')
```

So we have two options either a summary in the change log or a ` MIGRATION-5.0.0.md` file.
Do you have any preferences for situations like this?

mtrezza:
> By fixing these bugs and behaviors, we may unintentionally break their existing code instead of fixing it. 

Could you give an example where this may be breaking?

Nidal-Bakir:
Well, I don't have an example of a situation where this fix could break someone's existing code workaround. However, it's possible that someone has found a way to work around the bugs/incorrect behaviors in their code. If we are not responsible for the other workarounds, we can consider this a normal bug fix.

mtrezza:
> Well, I don't have an example of a situation where this fix could break someone's existing code workaround.

If we can't think of an example then let's assume that it's unlikely that there is such a relevant scenario. I've removed the `breaking` label.

> If this doesn't change the exposed behavior of the SDK then this is also not a `feat`. Does it fix a bug, improve performance / stability, or is it a `refactor`?

Nidal-Bakir:
> If we can't think of an example then let's assume that it's unlikely that there is such a relevant scenario. I've removed the breaking label.

Alright 

>If this doesn't change the exposed behavior of the SDK then this is also not a feat. Does it fix a bug, improve performance / stability, or is it a refactor?

It is a bug fix.

mtrezza:
Could you suggest a PR title that summarizes the bug that this PR fixes? In the description above the bugs are quite different from each other, so I'm unsure how to describe that in a single bug description.

Nidal-Bakir:
fix: Migrate the SDK to use `ParesOperation` class rather than `Map`s to perform parse-related operations 

What do you think?

mtrezza:
The PR title must describe the bug, not the fix or solution. The diff between fix and refactor is that refactor does not merit a new release. If it's a bug fix *and* a refactor then refactor takes precedence.

Nidal-Bakir:
fix: Return a `Map` rather than the actual data type `T` if an operation is performed on a key holds a `T` data type

is this good? or I should reword more

mtrezza:
This seems more like a description of the solution (return map instead of ...). How is the bug noticeable for the developer at the moment?

Nidal-Bakir:
> This seems more like a description of the solution 

not all it is the bug that every issue is pointing to. return Map rather than List or int or Relation after operation

mtrezza:
What should these example actually return?

Nidal-Bakir:
Yes 
`print(dietPlansObject.get('numKey'));` should return `3`

Nidal-Bakir:
> How do other Parse SDKs behave in this case?

Here is a screenshot using the JS SDK:
 
Screenshot from 2023-04-16 01-56-33
Screenshot from 2023-04-16 01-56-08


As you can see the first  `console.log` returns a `List` of the added elements and not what should be sent to the server. What should be sent to the server is `{__op: Add, objects: [3]}` _(Map)_ and the user should see `[3]` _(List)_. This SDK (Flutter SDK) returns to the user what will send to the server but the user should not see that and the user did not ask for it in the first place the user is asking for the current state of the list. 

The same thing for the second `console.log` returns a number `1` and that is what the value will be after the increment. Again the user asks for the current value of a key `'myNum'` not what the SDK will send to the server. the current SDK (Flutter SDK) will return what will send to the server(`{__op: Increment, amount: 1}`) rather than the value of the key.

mtrezza:
Makes sense. I understand now what you meant by this maybe being a breaking change for some even though it's a big fix.

It seems to me that there is no easy way to get from the op statement (like `{__op: Increment, amount: 1}`) to the actual value after the op statement is applied, because it contains only the delta, not the original value.

Nidal-Bakir:
> It may be possible that someone applied a workaround in which they store the original value somewhere and extract the delta value from the op statement

You've hit the nail on the head - that's precisely what worries me

mtrezza:
Any bug fix can potentially be a breaking change, so this alone wouldn't qualify for a major version release. But considering that this "behavior" has been around for so long, affecting all (?) of the atomic update methods, it sounds reasonable to consider publishing this as a major release.

I don't believe this requires a migration guide though, if returning the changed value instead of the op map is the only change. We could just add this in 1 sentence to the changelog in the `BREAKING CHANGES` section, what do you think?

Nidal-Bakir:
> affecting all (?) of the atomic update methods

Yes all the atomic update 

> I don't believe this requires a migration guide though, if returning the changed value instead of the op map is the only change. We could just add this in 1 sentence to the changelog in the BREAKING CHANGES section, what do you think?

We can do some explanation in the changelog. 

> Also, is there any other breaking change we want to merge together with this PR?

Yes, [bumping min Dart SDK]( https://github.com/parse-community/Parse-SDK-Flutter/issues/861#issuecomment-1484245270) because it is also a breaking change to do that.

mtrezza:
Sounds good!

Nidal-Bakir:
All right then.

I will do the next:

1. Add a changelog entry with a summary of the breaking change and bug fixes
2. Opens a new PR to bump the min Dart SDK version


also, should we use this as a title for this PR?

> Return a Map rather than the actual data type T if a parse-related operation is performed on a key that holds a T data type

or should we find another title?

mbfakourii:
@Nidal-Bakir Is it ready for the review?

Nidal-Bakir:
> is it ready for the review?

Yes go ahead

packages/dart/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: parse_server_sdk
 description: Dart plugin for Parse Server, (https://parseplatform.org), (https://back4app.com)
-version: 4.0.2
+version: 5.0.0
```
mbfakourii, line 3: Go directly to version 5? Isn't it better to have a beta version?

mbfakourii (dismissed):
Everything is great,tanks @Nidal-Bakir I think it is ready 🚀🚀

packages/dart/pubspec.yaml
Nidal-Bakir: Yes, we can do that.

@mtrezza  what do you think? should we publish this as beta?
Like this `5.0.0-beta.1`

[publishing-prereleases](https://dart.dev/tools/pub/publishing#publishing-prereleases)

mtrezza:
Should this be merged and a release published (breaking -> major release), or should this be released together with another PR?

Nidal-Bakir:
With other breaking PRs 
* #867

But let us merge this first so I can see the Changelog entry on the other PR and update it to state the min SDK bump

packages/dart/pubspec.yaml
mtrezza: We do not have this repository set up to be publish beta releases. If we publish beta releases on the same branch as stable releases we may run into trouble when adding auto-release later on.

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+         print(dietPlansObject.get('numKey')); 
+      ```
+      Previously this will print something like this `{__op: Increment, amount: 3}` rather than the result of the increment operation, which is `3`.
+      So now you will get the result of the operation (`3`) rather than a map of what will be sent to the server.
+
+
+### Bug Fixes
+
+* Calling `getRelation` on `ParseObject` after add/remove relation operation will throw: `type 'Map' is not a subtype of type 'ParseRelation<ParseObject>?' in type cast` ([#860](https://github.com/parse-community/Parse-SDK-Flutter/pull/860))
+
```
mtrezza, line 18:
```suggestion
## Unreleased

### BREAKING CHANGES

* Performing an atomic update on a key of a Parse Object now returns the prospective value, instead of a map of the operation that will be sent to the server; for example for a Parse Object `obj` with a key `count`, the atomic update `obj.setIncrement('count', 1);` previously returned the value `{__op: Increment, amount: 1}` but now returns the prospective result of the operation, which would be `1` if the key's previous value was `0`. ([#860](https://github.com/parse-community/Parse-SDK-Flutter/pull/860))

### Bug Fixes

* Setting atomic operation on Parse Object returns operation instead of prospective value ([#860](https://github.com/parse-community/Parse-SDK-Flutter/pull/860))

```

mtrezza (changes requested):
- Since this won't be released, there is no version 5.0.0, so it also can't be in the changelog yet.
- Reformatted the BREAKING CHANGES section to align with org-wide style.

mbfakourii:
@Nidal-Bakir 
A lot of values have been added. Does it not need to be added in the documentation?

Nidal-Bakir:
> A lot of values have been added. Does it not need to be added in the documentation?

No, we do not need to do that, because what this pull request is actually achieving is the default behavior, as in other SDKs. The previous behavior was actually a bug.

Nidal-Bakir:
The CI dry-run is failing because the change log does not mention the current version (5.0.0)

Nidal-Bakir:
> Since this won't be released, there is no version 5.0.0, so it also can't be in the changelog yet.

Then we will release 5.0.0 when we merge this PR #867 ?

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Unreleased
+
```
mtrezza, line 2:
```suggestion
## [5.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-5.0.0...dart-4.0.2) (UNRELEASED)

```

mtrezza (commented):
Maybe this will pass the dry run check.

mtrezza:
@Nidal-Bakir In my understanding https://github.com/parse-community/Parse-SDK-Flutter/pull/867 is the only other breaking PR we need to merge to make the 5.0.0 release. @mbfakourii Would you agree?

mbfakourii:
> @Nidal-Bakir In my understanding #867 is the only other breaking PR we need to merge to make the 5.0.0 release. @mbfakourii Would you agree?

Can't PR #867 be merged with this PR?

mtrezza:
Alright, CI passes, so we can merge this now?

Nidal-Bakir:
> Alright, CI passes, so we can merge this now?
 
Sure

mtrezza (approved):
Looks good!

mtrezza:
@parse-community/flutter-sdk-review Is this ready for merge?

packages/dart/CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
```
Nidal-Bakir, line 2: The (UNRELEASED) label is still present. we should remove it, right?

mtrezza: Good catch!

mtrezza (approved):

Nidal-Bakir:
We are ready to release Dart 5.0.0
Let us wait for the CI

Nidal-Bakir:
Alright, everything is good. lets move to the Flutter package

Nidal-Bakir:
Hay, Thank you for your kind words Iappreciate your feedback.

catalunha:
If I understood your problem correctly @juriSacchetta  automatically we only have complete data when it is a Pointer and we ask for it via includeObject. For objects in Relation fields we have to request the list separately via another manual query.

juriSacchetta:
Hello @Nidal-Bakir and @catalunha,

Thank you for your responses.

Upon further examination of my code, I have realized that the "stack overflow" error is caused by the usage of the `jsonencode` method instead of `toString` in `ParseBase`. I mistakenly attributed this error solely to the library version change, and I apologize for the confusion. However, I find it rather perplexing that such an error arises solely from the library version update. Can you offer any insights or suggestions as to why this might be happening?

Regarding the other problem, I will proceed to open an issue for it.

Thank you once again for your assistance.

maravilhosinga:
From this version there is a bug causing error in server side when doing some queries:

This is the error: 

`SyntaxError: Unexpected token } in JSON at position 335
    at JSON.parse (<anonymous>)
    at ClassesRouter.handleFind (/home/msinga/teego_server_devs/node_modules/parse-server/lib/Routers/ClassesRouter.js:29:25)
    at /home/msinga/teego_server_devs/node_modules/parse-server/lib/Routers/ClassesRouter.js:162:19
    at /home/msinga/teego_server_devs/node_modules/parse-server/lib/PromiseRouter.js:153:7
    at Layer.handle [as handle_request] (/home/msinga/teego_server_devs/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/msinga/teego_server_devs/node_modules/express/lib/router/route.js:144:13)
    at Route.dispatch (/home/msinga/teego_server_devs/node_modules/express/lib/router/route.js:114:3)
    at Layer.handle [as handle_request] (/home/msinga/teego_server_devs/node_modules/express/lib/router/layer.js:95:5)
    at /home/msinga/teego_server_devs/node_modules/express/lib/router/index.js:284:15
    at param (/home/msinga/teego_server_devs/node_modules/express/lib/router/index.js:365:14)
SyntaxError: Unexpected token } in JSON at position 335
    at JSON.parse (<anonymous>)
    at ClassesRouter.handleFind (/home/msinga/teego_server_devs/node_modules/parse-server/lib/Routers/ClassesRouter.js:29:25)
    at /home/msinga/teego_server_devs/node_modules/parse-server/lib/Routers/ClassesRouter.js:162:19
    at /home/msinga/teego_server_devs/node_modules/parse-server/lib/PromiseRouter.js:153:7
    at Layer.handle [as handle_request] (/home/msinga/teego_server_devs/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/msinga/teego_server_devs/node_modules/express/lib/router/route.js:144:13)
    at Route.dispatch (/home/msinga/teego_server_devs/node_modules/express/lib/router/route.js:114:3)
    at Layer.handle [as handle_request] (/home/msinga/teego_server_devs/node_modules/express/lib/router/layer.js:95:5)
    at /home/msinga/teego_server_devs/node_modules/express/lib/router/index.js:284:15
    at param (/home/msinga/teego_server_devs/node_modules/express/lib/router/index.js:365:14)`

Nidal-Bakir:
@maravilhosinga 

Please open an issue with your bug.

Issue #861, opened 2023-03-24
No named parameter with the name 'vsync'.


xfqwdsj:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description

When I built my app, I got an error: parse_server_sdk_flutter-4.0.0/lib/src/utils/parse_live_list.dart(276,9): error GC6690633: No named parameter with the name 'vsync'.

### Steps to reproduce

Building Windows app.

### Actual Outcome

The build fails with the error `No named parameter with the name 'vsync'`.

### Expected Outcome

The build was successful.

### Environment

Parse Flutter SDK
- SDK version: `4.0.0`
- Operating system version: `Windows 11 Build 22621.1413`

Server
- Parse Server version: `6.0.0`

### Logs

```
parse_server_sdk_flutter-4.0.0/lib/src/utils/parse_live_list.dart(276,9): error GC6690633: No named parameter with the name 'vsync'.
```

### Additional content

My Flutter version is `Channel flutter-3.9-candidate.15, 3.9.0-15.0.pre`, in this version, `vsync` parameter seems to be removed, I can manually delete the property specification of `vsync` in the downloaded SDK to fix this problem. This means that the solution to this problem can be as simple as removing a few lines of code from the SDK.

xfqwdsj:
> Can you test this on the Flutter stable channel? it seems that the bug is related to the Flutter framework, not the SDK

In the [documentation for Flutter's AnimatedSize Class](https://api.flutter.dev/flutter/widgets/AnimatedSize-class.html), the vsync property no longer exists, so this is an API change. I guess the SDK should no longer provide that property to AnimatedSize.

mtrezza:
What is the compatibility range of this SDK with Flutter?

xfqwdsj:
> What is the range of Flutter SDK versions that the Parse Flutter SDK currently supports?

[...]

PR #868, opened 2023-04-22
feat: Upgrade `parse_server_sdk` to `5.x.x`

Nidal-Bakir:

### Issue Description

Closes: #861

### TODOs before merging

- [ ] Add tests
- [x] A changelog entry


Nidal-Bakir:
The CI checks are falling because `parse_server_sdk: 5.0.0` has not been released yet.

Nidal-Bakir:
@mtrezza 
The last one!

mtrezza:
As you commented in https://github.com/parse-community/Parse-SDK-Flutter/pull/878#issuecomment-1546764559:

> But then if anyone uses Flutter SDK will be locked to Dart SDK 4.x.x.

Could we instead just say `>=4.0.0 <6.0.0` in #868?

Then #868 would not be a breaking change. I'm not sure if that makes sense given the Flutter / Dart co-dependency. But I guess we'd need to expand our CI to test the Parse Flutter SDK with the Parse Dart SDK 4.x and 5.x. Would it have any benefits for developers if we published a Parse Flutter SDK 5 that is compatible with Parse Dart SDK 4?

Nidal-Bakir:
I'm thinking about it. 
The thing is pub will never use the 4.0.0

Nidal-Bakir:
> Because every version of flutter_test from sdk depends on collection 1.17.0 and parse_server_sdk >=5.0.0 depends on collection ^1.17.1, flutter_test from sdk is incompatible with parse_server_sdk >=5.0.0.

And here is the answer to your question :)

mtrezza:
Right, I think it would only make sense if a developer would need to define both dependencies, the Parse Dart SDK *and* the Parse Flutter SDK in an app. Then they could in theory keep using their Parse Dart SDK 4.x while also using the Parse Flutter SDK 5.x. But is that even a realistic scenario?

Nidal-Bakir:
let me run some tests. that not good at all

mtrezza:
It seems strange because `flutter_test` is just the testing framework, right? So why would the testing framework influence the Parse SDKs compatibility. Maybe we need to change the testing framework then?

Nidal-Bakir:
If all the CIs resolve the dependency without any errors then we need to lower the collection version to : collection: ^1.16.0

Nidal-Bakir:
It seems strange because flutter_test is just the testing framework, right? So why would the testing framework influence the Parse SDKs compatibility. Maybe we need to change the testing framework then?

We can not change the testing framework

mtrezza:
> parse_server_sdk >=5.0.0 depends on collection ^1.17.1,

Can lower that dependency in the Parse Dart SDK and release a version 5.0.1?

Nidal-Bakir:
All the CIs resolve the dependency without any errors when using collection: ^1.16.0. on Dart SDK I test it on my PC also using Flutter 3.3.8 and 3.10. and all good when I changed to collection: ^1.16.0

We need to release Dart 5.0.1 updating the collection dependncy to be `collection: ^1.16.0.`

mtrezza:
Great, do you want to open a PR?

Nidal-Bakir:
Consider it done

Nidal-Bakir:
can you re-run the CI?

packages/flutter/CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
-## [5.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/flutter-4.0.0...flutter-5.0.0) (UNRELEASED)
+## [5.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/flutter-4.0.0...flutter-5.0.0)
```
mtrezza:
```suggestion
## [5.0.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/flutter-4.0.0...flutter-5.0.0) (2023-05-14)
```

Nidal-Bakir:
yeah, everything is green!

packages/flutter/pubspec.yaml
```diff
@@ -28,7 +28,6 @@ dependencies:
   sembast: ^3.4.0+6
   sembast_web: ^2.1.0+4
   path: ^1.8.2 # required for transitive use only
-  platform: ^3.1.0
```
mtrezza, line 31: Why has this been removed?

Nidal-Bakir: Not used.

mtrezza: Got it

mtrezza:
Can this be merged and we can then publish the Flutter 5 release?

Nidal-Bakir:
That's all. 

We should be good to go.

mtrezza:
Nice, great work!

Issue #873, opened 2023-05-11
Add support for Dart 3


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Current Limitation

To add official support for Dart 3 the CI needs to run tests using Dart 3. There is currently no Dart matrix and all tests run with Dart 2.19.x. In fact, it's not defined which Dart version should be used and the CI package may change the Dart version at any time.

### Feature / Enhancement Description

Add a Dart (and Flutter) version matrix to the CI to test against Dart (and Flutter) versions that are officially supported, according to our support policy currently discussed in #872.

Issue #872, opened 2023-05-09
Add Dart / Flutter framework support policy


mtrezza:
### New Feature / Enhancement Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Current Limitation

The Parse SDK is missing a policy that defines when support of a Dart / Flutter framework version should be dropped.

### Feature / Enhancement Description

Add a support policy depending on the LTS policy by the framework publishers, or if there is no such policy, according to community best practice.

mtrezza:
Dart and Fluter [don't seem to have a LTS policy](https://github.com/parse-community/Parse-SDK-Flutter/pull/867#issuecomment-1538986239). It seems like an easy concept to just not issue a LTS policy, but it doesn't address the issue LTS is trying to solve, which is helping developers (and dependency maintainers like us) to plan how much time they have until their app becomes incompatible with its frameworks.

Someone could open a PR today that makes the Parse SDK require the latest release of Dart and Flutter. We probably wouldn't merge it because it would force developers to make their app compatible with the latest Dart / Flutter release in order to use the latest version of the Parse SDK. That again will lower adoption of the latest Parse SDK release among developers, which means we'll get less feedback on newer releases, etc. It's the fundamental issue LTS is addressing.

Interestingly Dart doesn't follows semver, as the [docs](https://dart.dev/guides/language/evolution#language-version-numbers) say:
> Minor version numbers might introduce breaking changes.

I think we can assume that an increment of the lower bound of the supported Dart or Flutter framework is a breaking change as it makes our SDK incompatible with an app that uses a version below the new lower bound.

The [Dart release table](https://dart.dev/guides/language/evolution) shows that the release cycle varies by +/- 2 months. Approx. every 4 month we have a new breaking release.

The suggestion in https://github.com/parse-community/Parse-SDK-Flutter/pull/867 is to bump to Dart >= 2.17.0, which was released in 11 May 2022, basically a year ago. So maybe we can just establish a minimum support Dart version that is at least 1 year old. That would mean that a developer has 1 year to make their app compatible with Dart framework until their app may become incompatible with the Parse SDK.

mtrezza:
[...]

PR #874, opened 2023-05-11
ci: Add Dart / Flutter version matrix

mtrezza:

### Issue Description

CI is only testing against one (presumably the latest) dart framework version.

Closes: #873 

### Approach

- Add Dart SDK version matrix for testing of all currently officially supported Dart SDK versions.
- Add support policy by adding a compatibility table to the README of each package.

### DIFF
--- a/packages/flutter/README.md
+++ b/packages/flutter/README.md
@@ -6,8 +6,10 @@ This library gives you access to the powerful Parse Server backend from your Flu
 
 ---
 
+- [Compatibility](#compatibility)
 - [Getting Started](#getting-started)
     - [Web support](#web-support)
+    - [Desktop Support (macOS)](#desktop-support-macos)
     - [Network client](#network-client)
 - [Objects](#objects)
 - [Custom Objects](#custom-objects)
@@ -36,6 +38,17 @@ This library gives you access to the powerful Parse Server backend from your Flu
 - [Other Features](#other-features)
 
 ---
+
+## Compatibility
+
+The Parse Flutter SDK is continuously tested with the most recent release of the Flutter framework to ensure compatibility. To give developers time to upgrade their app to the newest Flutter framework, previous Flutter framework releases are supported for at least 1 year after their [release date](https://docs.flutter.dev/release/archive?tab=linux).
+
+| Version      | End of Support | Compatible                                   |
+|--------------|----------------|----------------------------------------------|
+| Flutter 3.10 | May 2024       | ✅ Yes                                        |
+| Flutter 3.7  | Apr 2024       | ✅ Yes                                        |
+| Flutter 3.3  | Jan 2024       | ✅ Yes                                        |
+| Flutter 3.0  | Jul 2023       | ❌ No (Parse Dart SDK requires Dart >=2.18.0) |
 
 ## Getting Started
 
--- a/packages/flutter/README.md
+++ b/packages/flutter/README.md
@@ -45,9 +45,9 @@ The Parse Flutter SDK is continuously tested with the most recent release of the
 
 | Version      | End of Support | Compatible                                   |
 |--------------|----------------|----------------------------------------------|
-| Flutter 3.10 | May 2024       | ✅ Yes                                        |
+| Flutter 3.10 | May 2024       | ❌ No                                         |
 | Flutter 3.7  | Apr 2024       | ✅ Yes                                        |
-| Flutter 3.3  | Jan 2024       | ✅ Yes                                        |
+| Flutter 3.3  | Jan 2024       | ❌ No                                         |
 | Flutter 3.0  | Jul 2023       | ❌ No (Parse Dart SDK requires Dart >=2.18.0) |
 
 ## Getting Started
--- a/packages/flutter/README.md
+++ b/packages/flutter/README.md
@@ -41,7 +41,7 @@ This library gives you access to the powerful Parse Server backend from your Flu
 
 ## Compatibility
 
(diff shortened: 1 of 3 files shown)


Nidal-Bakir:
I see that we need to update the examples to support null safety.

mtrezza:
I've added the Flutter version matrix now as well. According to our support policy (tbd), we would test with Flutter 3.10 (latest), 3.7, 3.3, 3.0. But only Flutter 3.7 seems to succeed, others failing.

mtrezza:
The Flutter 3.0 test fails with:

> ERR : The current Dart SDK version is 2.17.6.
>     | 
>     | Because parse_server_sdk depends on collection >=1.17.1 which requires SDK version >=2.18.0 <3.0.0, version solving failed.
> FINE: Exception type: SolveFailure

That's expected because Flutter 3.0.5 [requires](https://docs.flutter.dev/release/archive?tab=linux) Dart 2.17.6, but as we no know the Parse Dart SDK requires at least Dart 2.18.0. Just like [expected](https://github.com/parse-community/Parse-SDK-Flutter/issues/872#issuecomment-1540016654):

> Flutter version at least 1 year old, or the Flutter version that requires the lowest supported Dart version by the Parse Dart SDK, whatever is higher.

So the Parse Flutter SDK can only support Flutter >=3.3 and we can actually remove the Flutter 3.0 test from the CI.

Nidal-Bakir:
Flutter 3.10.0 and Dart 3 require full sound null safety but the examples are not null safe


Flutter 3.3
>Error: No named parameter with the name 'exclusive'.
I think its because the named parameter 'exclusive' is not present in this version of Flutter/Dart

mtrezza:
And do you know why Flutter lint fails in this PR?

Nidal-Bakir:
Maybe because of the deprecated "format" command

mtrezza:
That worked.

Flutter code analysis fails with

> The current Dart SDK (3.0.0) only supports null safety.

So the code analysis seems to depend on the Dart / Flutter version. Should we do the code analysis as part of the matrix tests instead of as a separate test? Then the code analysis would run specifically with each Dart / Flutter version.

mtrezza:
> If this is the case then we can not use Flutter 3.3 and we need to set a constraint for Flutter
> like this flutter: >=3.7.0 <=3.10.0 but Flutter 3.7.0 comes with Dart 2.19.0 
> so we can say that we support  sdk: >=2.19.0 <=3.0.0

Do you mean there is no way to set a Flutter constraint in the pubspec, but it can be done indirectly by setting the Dart constraint?

```
environment:
  sdk: ">=2.12.0 <3.0.0"
```

Nidal-Bakir:
That will not fix the error 
will pop out on the latest flutter/dart 3.10.0/3

We need to update the examples to support null safety.

mtrezza:
The version range was just an example, I mean is there no way to set the flutter version range in the pubspec? Can we only set the dart range?

Nidal-Bakir:
> Do you mean there is no way to set a Flutter constraint in the pubspec, but it can be done indirectly by setting the Dart constraint?

We can set both in the Flutter package
like this:
```
environment:
  sdk: ">=2.19.0 <=3.0.0"
  flutter: ">=3.7.0 <=3.10.0" 
```

but on the Dart package we can only set the Dart constraint

```
environment:
  sdk: ">=2.19.0 <=3.0.0"
```

Nidal-Bakir:
could remove the tests for any` ci / Test Dart  <2.18`  no need for them

mtrezza:
This PR is only to add the missing CI tests. We'll remove the tests in the other PR where we change the SDK version range.

mtrezza:
I have added the publish dry-run and lint to the matrix, because both of them seem to depend on the dart / flutter version.

Nidal-Bakir:
I do not see any issue with that.

mtrezza:
Can we merge this?

packages/flutter/README.md
```diff
@@ -37,6 +39,17 @@ This library gives you access to the powerful Parse Server backend from your Flu
...
 
+## Compatibility
+
+The Parse Flutter SDK is continuously tested with the most recent release of the Flutter framework to ensure compatibility. To give developers time to upgrade their app to the newest Flutter framework, previous Flutter framework releases are supported for at least 1 year after their [release date](https://docs.flutter.dev/release/archive?tab=linux). The Parse Flutter SDK depends on the Parse Dart SDK which may require a higher Dart framework version than the Flutter framework version, in which case the Flutter framework version cannot be supported even though its release date may have been less than a year ago.
+
+| Version      | End of Support | Compatible                                   |
+|--------------|----------------|----------------------------------------------|
+| Flutter 3.10 | May 2024       | ✅ Yes                                        |
+| Flutter 3.7  | Apr 2024       | ✅ Yes                                        |
+| Flutter 3.3  | Jan 2024       | ✅ Yes                                        |
```
Nidal-Bakir, line 50: We will change this letter before we publish version 5.0.0, right? because Flutter 3.3 probably will not work

The same for Dart 2.18 & Dart 2.17 probably will not work

mtrezza: Yes, we can change this as needed before release.

The current Flutter table looks like this:
image

Because the CI passes only for Flutter 3.7 so far :-/

Nidal-Bakir (approved):

packages/flutter/README.md
Nidal-Bakir: Yes, in current situation is accurate

Nidal-Bakir:
> Can we merge this?

We are good to go!

Issue #882, opened 2023-05-14
Improve dart pub points scoring on pub.dev


mtrezza:
### New Issue Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description

pub.dev [pub points](https://pub.dev/packages/parse_server_sdk/score) analysis shows:

1. Maybe we need to remove the upper bound `<3.1.0` because pub.dev doesn't like clear compatibility specifications:

> Unable to run pub upgrade.
> The current Dart SDK version is 3.1.0-63.1.beta.
> 
> Because parse_server_sdk requires SDK version >=2.18.0 <3.1.0, version solving failed.
> 
> You can try the following suggestion to make the pubspec resolve:
> * Try using the Dart SDK version: 3.0.0. See https://dart.dev/get-dart.

Also interesting, it's using a beta version of dart 3.1 (there has been no stable 3.1 release yet) to determine compatibility. An odd choice to let a beta version determine the public scoring of a package.

2. The package description is too short:

> Add more detail to the description field of pubspec.yaml. Use 60 to 180 characters to describe the package, what it does, and its target use case.

3. No example:

> 0/10 points: Package has an example
>  No example found.

mbfakourii:
@mtrezza 
Should I make a PR for the Flutter version as well? Or create another PR?

PR #883, opened 2023-05-14
refactor: Increment Dart SDK compatibility to `<4.0.0`; add example project

mbfakourii:

### Issue Description
Add example for dart pakage
Add screenshot
Increase the sdk environment range to 4 (base ([flutter packages](https://github.com/material-foundation/flutter-packages/commit/e7bd3c4c42df3b60b34c6c89c8de0687f535b6c1)))
Increase the description between 60 and 180 characters

Closes: #882

### Approach
n/a

### TODOs before merging

- [ ] Add tests
- [ ] A changelog entry


packages/dart/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: parse_server_sdk
-description: The Dart SDK for Parse Platform (https://parseplatform.org)
+description: The Dart SDK for Parse Platform, based on Parse Server (https://parseplatform.org).
 version: 5.1.0
 homepage: https://github.com/parse-community/Parse-SDK-Flutter
 
 environment:
-  sdk: ">=2.18.0 <3.1.0"
+  sdk: ">=2.18.0 <4.0.0"
```
mtrezza, line 7: This goes against how dart works, which is that minor versions may contain breaking changes. Setting <4.0.0 is effectively an open range, so we can as well just remove the upper bound altogether. It seems to be a conceptual bug in dart.

I think we have these options:
- a) Stay correct about the compatibility range, which is _in reality_ `<3.1.0`.
- b) Use an incorrect compatibility range (the current SDK release may not be compatible with dart 3.1.0) just to please the flawed pub.dev score points determination.
- c) Add the beta version of the next dart version to the CI and claim a potentially less incorrect range of `<=3.1.0`.

While (c) sounds a bit complicated, it may actually help us to prepare for the next minor release or dart, so we can prepare the SDK if needed before stable release.

mbfakourii: @mtrezza 

Why don't we proceed in the way of famous packages? See [this page](https://github.com/material-foundation/flutter-packages/commit/e7bd3c4c42df3b60b34c6c89c8de0687f535b6c1)

mtrezza: That would be option (b). It would still be good if we understood why we are making that decision.

mbfakourii: @mtrezza What do you think we should do?

what is your opinion ?
@Nidal-Bakir

mbfakourii: @Nidal-Bakir 
what is your opinion?

Nidal-Bakir: I think the best way to handle this is by applying the two options (b) and (c).

By testing the SDK against the beta version of Flutter/Dart and setting the upper bound to `<4.0.0`. By doing this, we will be able to detect any issues with the new beta version of Flutter/Dart immediately.

mtrezza: We could try that. (b) fixes analyzer warning and (c) helps to prepare for the next release. If updating the CI job to the latest beta version for (c) turns out to be too much work we can rethink that. All on board?

Nidal-Bakir: Yes!

packages/dart/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: parse_server_sdk
-description: The Dart SDK for Parse Platform (https://parseplatform.org)
+description: The Dart SDK for Parse Platform, based on Parse Server (https://parseplatform.org).
```
mtrezza, line 2:
```suggestion
description: The Dart SDK to connect to Parse Server. Build your apps faster with Parse Platform, the complete application stack.
```

mtrezza (changes requested):

packages/dart/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: parse_server_sdk
-description: The Dart SDK for Parse Platform (https://parseplatform.org)
+description: The Dart SDK to connect to Parse Server. Build your apps faster with Parse Platform, the complete application stack.
 version: 5.1.0
```
mtrezza, line 3: I think we need to publish a release for this PR, to get the new score on pub.dev. Alternatively we could not publish a release, merge this as `refactor` and just keep the bad score until the next dart release, but it wouldn't hurt to get a better score right away.

```suggestion
version: 5.2.0
```

packages/dart/pubspec.yaml
mbfakourii: > We could try that. (b) fixes the analyzer warning and helps with the score and (c) helps to prepare for the next release. If updating the CI job to the latest beta version for (c) turns out to be too much work we can rethink that. All on board?

It seems good

packages/dart/pubspec.yaml
mbfakourii: I don't think we need to give a new version for this PR, only a ```refactor``` will be better

mtrezza: Alright

mtrezza (approved):
Looks good!

packages/dart/pubspec.yaml
mtrezza: Alright

Issue #149, opened 2019-04-11
offline capabilities and auto-syncing


marqueeeeeee:
Does this support offline saving and auto-sync when network is available? similar to saveEventually from parse sdk and firebase db? Thank you!

rodrigosmarques:
Hi @marqueeeeeee,
this feature is not available yet.

frsisalima:
Hello, is the implementation of this feature planned? it would be great to have offline support

monajafi:
we can implement saveEventually in android by caching saves if there is no connectivity then with help of workmanager plugin which is a wrapper around android WorkManager we can sync the cached saves when the device connects to internet
This is easy in android but I'm trying to figure out how to implement in iOS because the background jobs are restricted in iOS we have just background fetch.
One point to have in mind about saveEventually is that there is no callback after saving 
I have an app which is developed using native sdk If the server is down and device get connected to internet the save would fail

phillwiggins:
Hey Noida

We should stick to keeping it pure Dart.

We simply need a connection manager and a save pool of objects.

monajafi:
One important thing is that after successful save the save pool should be cleared not just after gaining connectivity and saving without considering the save result, which is a behaviour in native sdks

phillwiggins:
100% correct logic. We only clear the pool, once confirmed successfully
saved.

I was looking at this but in the past few months been too busy to implement.

cadaniel:
Would love to see this feature. It's basically all I'm missing from what I need to replace firebase in my app (hopefully before launching)

Is there some engineering docs for this that I could help contribute?

RodrigoSMarques:
The project is using SEMBAST which is also a NoSQL database written in Dart.
The sample application is using SEMBAST to do some offline resources. Take a look at the sample project.

You can study the source of Parse SDK on Android to see how it works. This is how I contributed to this plugin.

Sun3:
I also would love to see saveEventually implemented for both Android and iOS.

allanwolski:
It's very important to have offline support for our application.

visual-velocity:
> It's very important to have offline support for our application.

I agreed

paulocagol:
Hello, any progress?

RodrigoSMarques:
No progress.

RodrigoSMarques:
No progress.

paulocagol:
Okay, thanks for the feedback.

2shrestha22:
[...]

Issue #194, opened 2019-06-12
Optimize ParseObject


yulingtianxia:
### Creating a ParseObject

- [x] ParseObject()

### Managing Object Properties

- [x] parseClassName //(Need refactor)
- [x] objectId
- [x] updatedAt
- [x] createdAt
- [x] ACL

### Accessors

`remove` conflicts with existing methods.

- [x] get
- [x] set
- [ ] remove
- [x] getRelation
- [ ] revertAll
- [ ] revert

### Array Accessors

`remove` conflicts with existing methods.

- [x] add
- [x] addAll
- [x] addUnique
- [x] addAllUnique
- [x] remove
- [x] removeAll

### Increment

- [x] increment //(Need refactor)
- [x] decrement //(Need refactor or delete)

### Saving Objects

- [x] save //(Need refactor)
- [ ] saveEventually

### Saving Many Objects(Static Method)

- [ ] saveAll

### Deleting Many Objects(Static Method)

- [ ] deleteAll

### Getting Many Objects(Static Method)

- [ ] fetchAll //(Has replacement `getAll`，Need refactor)
- [ ] fetchAllIfNeeded

### Getting an Object

- [ ] dataAvailable
- [ ] fetch //(Has replacement `getObject`，Need refactor)
- [ ] fetchIfNeeded

### Fetching From Local Datastore

- [ ] fetchFromLocalDatastore

### Deleting an Object

- [x] delete
- [ ] deleteEventually

### Dirtiness

- [ ] dirty({String key})

### Pinning

- [ ] pin({String name, dynamic callback}) //(Need refactor)

### Pinning Many Objects(Static Method)

- [ ] pinAll({String name, dynamic callback}) 

### Unpinning

- [ ] unpin({String name, dynamic callback}) //(Need refactor)

### Unpinning Many Objects(Static Method)

- [ ] unpinAll({bool recursively, String name, dynamic callback}) 

### Methods for Subclasses(Static Method)

- [ ] object
- [ ] objectWithoutDataWithObjectId
- [ ] registerSubclass
- [ ] query()

maaeps:
registerSubclass would be great.

Maybe changing fomJson in ParseBase could do the trick.
Any ideas?

maaeps:
Maybe getting close to the official repos is not that good.
There are some cases where having the android like subclass system is not the best.

maaeps:
The registerSubclass feature got merged with #368

PR #911, opened 2023-05-19
feat: Add `saveEventually` and `deleteEventually` in `ParseObject`

mbfakourii:
## Pull Request

## Issue

Closes: #149 and #194

## Approach

In this feature, Functions `saveEventually` and `deleteEventually` in `ParseObject` were added

---> in `saveEventually` 
At first, normal ```save``` command is entered in parseObject, if a ```NetworkError``` is encountered, this object is saved in a list with a key with the help of ParseCoreData.
In our collection, we have a list of objects in ```ParseCoreData``` that are in ```NetworkError```.

---> in `deleteEventually` 
At first, normal ```delete``` command is entered in parseObject, if a ```NetworkError``` is encountered, this object is saved in a list with a key with the help of ParseCoreData.
In our collection, we have a list of objects in ```ParseCoreData``` that are in ```NetworkError```.

---> in ```await ParseObject.submitEventually()```
With the help of the function ```await ParseObject.submitEventually()```  (which is debatable how to use it) we checks the error objcets

In this function, we first get the error parseObjects from ```ParseCoreData```, then we send this list to ```_saveChildren```. In this function, these objects are sent to the server in the form of chunks, and finally, we empty the ```ParseCoreData``` list.

In the following, we also take the list of ```parseObjects``` that must be deleted from ```ParseCoreData``` and try to delete them, and then we delete the successfully deleted parseObjects from the ```ParseCoreData``` list.

Please check these two functions thoroughly and give suggestions, this PR needs discussion.
and please check the algorithm, is this method good?

## Tasks

- [x] Add tests


mtrezza:
> Please check these two functions thoroughly and give suggestions, this PR needs discussion.

Changed to draft

mbfakourii:
@mtrezza @Nidal-Bakir 
Do you think the implemented method is appropriate?

Nidal-Bakir:
I suggest looking at what other SDKs do and doing the same. There is the concept of a Queue so the objects get saved in order. 

Do other SDKs save the unsaved Objects even when the app is closed(destroyed) using native services (android and iOS) when the connection is restored for example?

mbfakourii:
> I suggest looking at what other SDKs do and doing the same. There is the concept of a Queue so the objects get saved in order.
> 
> Do other SDKs save the unsaved Objects even when the app is closed(destroyed) using native services (android and iOS) when the connection is restored for example?

I think the queue issue happens in [dio](https://pub.dev/documentation/dio/latest/dio/QueuedInterceptor-class.html) or Http and we don't need to create this queue.

According to the ios documentation

[saveEventually](https://parseplatform.org/Parse-SDK-iOS-OSX/api/Classes/PFObject.html#/c:objc(cs)PFObject(im)saveEventually)
```
Saves this object to the server at some unspecified time in the future, even if Parse is currently inaccessible.

Use this when you may not have a solid network connection, and don’t need to know when the save completes. If there is some problem with the object such that it can’t be saved, it will be silently discarded.

Objects saved with this method will be stored locally in an on-disk cache until they can be delivered to Parse. They will be sent immediately if possible. Otherwise, they will be sent the next time a network connection is available. Objects saved this way will persist even after the app is closed, in which case they will be sent the next time the app is opened. If more than 10MB of data is waiting to be sent, subsequent calls to -saveEventually will cause old saves to be silently discarded until the connection can be re-established, and the queued objects can be saved.
```

Based on what I understand, the internal structure of Android is implemented with this concept. If possible, take a look at the [Android structure](https://github.com/parse-community/Parse-SDK-Android/blob/2bf43513973277558a00e55ad96d0f5b33baeb5b/parse/src/main/java/com/parse/ParseObject.java#L2339).

The concept that I didn't understand was where to double check and send the saved it objects?

I wrote the codes of this section in the ```await ParseObject.submitEventually()``` function! And I allow the user to use it wherever they need. But not sure about that !

mbfakourii:
@parse-community/android-sdk and @parse-community/ios-sdk

It is possible for you to participate in this discussion

packages/dart/lib/src/objects/parse_object.dart
```diff
@@ -677,4 +722,89 @@ class ParseObject extends ParseBase implements ParseCloneable {
       return this;
     }
   }
+
+  static Future<ParseResponse?> submitEventually(
```
Nidal-Bakir, line 726: who should call this faction the end-user or the SDK itself when initialized?

Nidal-Bakir (commented):
Overall everything is good.

I have some thoughts:
- What will happen if the current user logs out?
  -  should we clear the cache? or could you give the end user the option to decide what should happen?
- should we give the end user control over the cached objects that need to be saved/deleted?

packages/dart/lib/src/objects/parse_object.dart
```diff
@@ -677,4 +722,89 @@ class ParseObject extends ParseBase implements ParseCloneable {
...
+    List<String>? listSaves =
+        await coreStore.getStringList(keyParseStoreObjects);
+
+    if (listSaves != null) {
+      List<ParseObject> parseObjectList = [];
+      for (var element in listSaves) {
+        // decode json
+        dynamic object = json.decode(element);
+        parseObjectList
+            .add(ParseObject(object[keyVarClassName]).fromJson(object));
```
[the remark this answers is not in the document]
mbfakourii: ClassName is not used in the fromJson function, I had to write it like this

mbfakourii:
> Overall everything is good.
> 
> I have some thoughts:
> 
> * What will happen if the current user logs out?
>   
>   * should we clear the cache? or could you give the end user the option to decide what should happen?
> * should we give the end user control over the cached objects that need to be saved/deleted?

At the moment, the caches are not deleted, but I think a tool should be written to delete the cache

Regarding the second question, I did not see anything in other SDKs. Do you have any suggestions for this? What control do you mean?

Nidal-Bakir:
> At the moment, the caches are not deleted, but I think a function should be written to delete the cache

Got it.


> Regarding the second question, I did not see anything in other SDKs. Do you have any suggestions for this? What control do you mean?

I mean what if the user wants to clear the currently cached objects for saving/deleting and start over?
Should we give expose such a feature?

mbfakourii:
> I mean what if the user wants to clear the currently cached objects for saving/deleting and start over? Should we expose such a feature?
> 
> Or in general, should we expose the current cache state so the end user can alter it?
> 
> If yes we should consider a saving mechanism like functions to be invoked. for example:
> 
> * onBeforStartSaving()
> * shouldSave()
> * shouldDelete()
> * onBeforSave()
> * onBeforDelete()
> * onAfterSave()
> * onAfterDelete()
> * onAllSaved()
> * onAllDeleted()
> 
> something like that

I don't think this is necessary. I haven't seen anything about this in other SDKs. Do you have a reason why it needs to be added?

Nidal-Bakir:
No, I'm just wounding if we need to give the end user finer control over the eventual save in general. If the other SDKs do not do that we are good.

So currently, we need to discuss how the submitEventually function should be called and what will happen when the user logs out when there are some unsaved objects.

mbfakourii:
> > Do you have a reason why it needs to be added?
> 
> No, I'm just wounding if we need to give the end user finer control over the eventual save in general. If the other SDKs do not do that we are good.
> 
> So currently, we need to discuss how the `submitEventually` function should be called and what will happen when the user logs out when there are some unsaved objects.

Do you think we should leave this function to the user or place it somewhere like the `save()` function?

mtrezza:
What's missing in this PR to get it merged?

mbfakourii:
> What's missing in this PR to get it merged?

The code is finished, the tests must be write

mbfakourii:
The error related to [`Test Flutter beta`](https://github.com/parse-community/Parse-SDK-Flutter/actions/runs/6397441480/job/17365365740?pr=911) has been resolved in [this](https://github.com/parse-community/Parse-SDK-Flutter/pull/969) PR

packages/dart/lib/src/objects/parse_object.dart
```diff
@@ -162,7 +162,7 @@ class ParseObject extends ParseBase implements ParseCloneable {
       return response;
     } else {
       // if have network connection error
-      if ((response.error?.message ?? "") == "NetworkError") {
+      if ((response.error?.message ?? "").contains("NetworkError")) {
```
mtrezza, line 165: Why is this parsing an error message, is there no error code to compare?

mbfakourii: I changed it, please check again

mbfakourii:
@mtrezza 

Done 😃

packages/dart/lib/src/objects/parse_object.dart
```diff
@@ -271,7 +290,7 @@ class ParseObject extends ParseBase implements ParseCloneable {
         final ParseResponse response = await batchRequest(
           requests,
           chunk,
-          client: _client,
+          client: client,
```
mtrezza, line 293:
```suggestion
          client,
```

packages/dart/lib/parse_server_sdk.dart
```diff
@@ -27,62 +27,116 @@ export 'src/network/parse_dio_client.dart';
 export 'src/network/parse_http_client.dart';
 
 part 'src/base/parse_constants.dart';
+
```
mtrezza, line 30: Are these empty lines an automatic lint change or were they added by accident?

mtrezza:
Note for review: The Flutter test is failing, hence we need to merge https://github.com/parse-community/Parse-SDK-Flutter/pull/969 first before merging this PR.

packages/dart/lib/src/objects/parse_object.dart
mbfakourii: This suggestion is not correct, it creates a conflict.

packages/dart/lib/parse_server_sdk.dart
mbfakourii: This has been generated using [auto code formatting](https://docs.flutter.dev/tools/formatting#automatically-formatting-code-in-android-studio-and-intellij) in Dart.

mtrezza (commented):
Could you check whether I've resolved the conflicts correctly? Then I believe this can be merged.

mbfakourii:
> Could you check whether I've resolved the conflicts correctly? Then I believe this can be merged.

I checked, there was no problem

packages/dart/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## [6.2.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-6.1.0...dart-6.2.0) (2023-10-17)
```
```suggestion
## [6.2.0](https://github.com/parse-community/Parse-SDK-Flutter/compare/dart-6.1.0...dart-6.2.0) (2023-10-18)
```

mtrezza (approved):

PR #934, opened 2023-06-07
refactor: Update package_info_plus

mbfakourii:
## Pull Request

## Issue

Closes: #821

## Approach
Update package_info_plus to ```package_info_plus: ^4.0.2```

## Tasks

- [ ] Add tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Analyze code
         run: dart analyze packages/dart --fatal-infos
       - name: Lint
-        run: dart format --output=none --set-exit-if-changed packages/dart
+        run: dart format --output=show --set-exit-if-changed packages/dart
       - name: Publish dry run
         run: cd packages/dart && dart pub publish --dry-run
       - name: Run tests
@@ -127,7 +127,7 @@ jobs:
       - name: Analyze code
         run: flutter analyze packages/flutter --fatal-infos
       - name: Lint
-        run: dart format --output=none --set-exit-if-changed packages/flutter
+        run: dart format --output=show --set-exit-if-changed packages/flutter
       - name: Publish dry run
         run: cd packages/flutter && dart pub publish --dry-run
       - name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,6 @@ jobs:
           - name: Dart 2.18
             os: ubuntu-latest
             sdk: 2.18.7
-          - name: Dart beta
-            os: ubuntu-latest
-            sdk: beta
       fail-fast: false
     name: Test ${{ matrix.name }}
     steps:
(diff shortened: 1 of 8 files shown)


mtrezza (changes requested):
CI tests failing...

mtrezza:
Re-running CI to get all tests to pass

mbfakourii:
> It seems that the CI is always failing, and it's not just the coverage upload. Could you take a look?

Yes, Sure

mbfakourii:
The last error related to the [code format](https://github.com/parse-community/Parse-SDK-Flutter/actions/runs/5222554583/jobs/9428184149) is not a problem in my system, but it is a problem in CI!

image

mtrezza:
I think the issue is that the linter is part of the dart framework. So if the Dart beta version enforces a new lint rule, the lint check for that framework version will fail. That's why it's only failing for "Dart beta".

mbfakourii:
@mtrezza 
Why should the beta version be checked?

mtrezza:
It fails `dart format`, which I assume is because the linter changed in that dart version.

.github/workflows/ci.yml
```diff
@@ -33,9 +33,6 @@ jobs:
           - name: Dart 2.18
             os: ubuntu-latest
             sdk: 2.18.7
-          - name: Dart beta
```
mtrezza, line 36: Why removed?

mbfakourii:
Beta removed from ci

mtrezza:
Why removed?

mbfakourii:
> Why removed? Adding beta was debated thoroughly in a previous PR and we deiced to add it. If we now remove it we should find arguments that outweigh the other arguments.
> 
> I think we'd need to:
> 
> 1. Find out what exactly the linter is complaining about.
> 2. If we can't fix it, remove linting for beta, instead of removing beta altogether.
> 3. If we can't removing linting for beta, decide whether we should remove beta.

I was testing and returned it

mtrezza:
Do you think you could run `dart format` locally with a beta version to find out what the 1 issue is the linter corrects?

mbfakourii:
> Do you think you could run `dart format` locally with the dart beta version to find out what the 1 issue is the linter corrects?

I will test

mbfakourii:
@mtrezza 
It has problems from version 2.18 to 3. I will roll back to the previous version. I think it is better if there is a problem in the beta!

mtrezza:
Is there no way we can slightly refactor the code so it passes both?

mbfakourii:
> Is there no way we can slightly refactor the code so it passes both?

I will check

mbfakourii:
fixed

mtrezza (approved):
Well done! So we can keep the beat versions in the CI; is this ready for merge?

mbfakourii:
> Well done! So we can keep the beat versions in the CI; is this ready for merge?

Yes, it is ready to merge

Issue #958, opened 2023-08-01
Version 5.1.2 compile problems


hello96:
### New Issue Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [x ] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description
It's not possible to build a project when the SDK is installed. I think it should be something related to latest Dart versions, but I'm not sure.

### Steps to reproduce
1. Create an empty flutter application;
2. Import latest parse sdk version (currently 5.1.2);
3. Call Parse().initialize()

### Actual Outcome
![image](https://github.com/parse-community/Parse-SDK-Flutter/assets/31999705/f55039ab-28e1-41b5-8806-79f74f9e7a31)

### Expected Outcome
To work :)

### Environment
Parse Flutter SDK
- SDK version: `5.1.2`
- Flutter version: `3.10.5`
- Dart version: `3.0.5`
- Operating system version: `Windows 11 and MacOS M1`

Server
- Parse Server version: `4.5.0`

### Logs
![image](https://github.com/parse-community/Parse-SDK-Flutter/assets/31999705/f55039ab-28e1-41b5-8806-79f74f9e7a31)

GIovaB6:
Yeah, i've the same problem

codakkk:
I'm encountering the same issue, right now

mbfakourii:
It seems to be from `Dio`, I will check

Piyush-e7:
Yeah, I have the same problem

mbfakourii:
The problem has been solved on the [dio side](https://github.com/cfug/dio/pull/1916). I am waiting for the new version of dio. Please use the old versions until the new version thanks

mbfakourii:
@hello96 @GIovaB6 @codakkk @Piyush-e7 

Has the problem been solved?

mbfakourii:
Bug Fixed

PR #956, opened 2023-07-30
refactor: Improve pubspecs files and example projects

mbfakourii:
## Pull Request

- Report security issues [confidentially](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- Any contribution is under this [license](https://github.com/parse-community/Parse-SDK-Flutter/blob/master/LICENSE).
- Link this pull request to an [issue](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

## Issue
Closes: #957

## Approach
Improved pubspecs files base [dart site](https://dart.dev/tools/pub/pubspec) and example projects

## Tasks

- [ ] Add tests
- [ ] Add changes to documentation (guides, repository pages, code comments)

PR #959, opened 2023-08-03
fix: Update dio

mbfakourii:
## Pull Request

## Issue

Closes: #958

## Approach
Update dio to ```dio: ^5.3.2```

## Tasks

- [ ] Add tests

### DIFF
--- a/packages/dart/pubspec.yaml
+++ b/packages/dart/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
 
 dependencies:
   # Networking
-  dio: ^5.2.0+1
+  dio: ^5.3.2
   http: ^0.13.5
   web_socket_channel: ^2.4.0
 
--- a/packages/dart/pubspec.yaml
+++ b/packages/dart/pubspec.yaml
@@ -8,7 +8,7 @@ environment:
 
 dependencies:
   # Networking
-  dio: ^5.3.2
+  dio: ^5.2.0+1
   http: ^0.13.5
   web_socket_channel: ^2.4.0
 


mtrezza (approved):
Looks good, could you write a PR title that describes the issue (not the solution)

mbfakourii:
> Looks good, could you write a PR title that describes the issue (not the solution)

I have no other idea, only dio was updated, the reason was a bug in the dio package

mtrezza:
Why did we have to update the dio package, if we are using a range operator anyway?

```diff
- dio: ^5.2.0+1
+ dio: ^5.3.2
```

If I understand this correctly, as soon as a new version (non-major) version is released, it will be updated anyway?

mbfakourii:
> Why do we have to update the dio package, if we are using a range operator anyway?
> 
> ```diff
> - dio: ^5.2.0+1
> + dio: ^5.3.2
> ```
> 
> If I understand this correctly, `^5.2.0+1` will use the latest dio version <6.0.0 as soon as a newer version is released, right?

Because of [these issues](https://github.com/parse-community/Parse-SDK-Flutter/issues/958), we have to give a new version quickly because the current version [cannot be used](https://github.com/parse-community/Parse-SDK-Flutter/issues/958#issuecomment-1661922253).

I checked the dio package and did not see anything from version 6

mtrezza:
But why do we have to update our package? In my understanding, the Parse Flutter SDK will always install the latest release of dio <6, because the dependency is `^5.2.0+1`. So even if we don't update the package, Parse Flutter SDK should use `5.3.2`, not `5.2.0`.

mbfakourii:
Yes, I think the problem was solved with [Revert](https://github.com/cfug/dio/pull/1916) on the dio side!
I am waiting for the answer of [those who have problems](https://github.com/parse-community/Parse-SDK-Flutter/issues/958#issuecomment-1666484787)

mtrezza:
I think we should see this also in our [CI](https://github.com/parse-community/Parse-SDK-Flutter/pull/956#issuecomment-1666501726).

mbfakourii:
> I think we should see this also in our [CI](https://github.com/parse-community/Parse-SDK-Flutter/pull/956#issuecomment-1666501726). In fact the CI previously failed and now passes. In my opinion this PR can be closed, as well as the related issue, because it was a dio issue, not a Parse issue.

sure

Issue #966, opened 2023-09-25
Update Dependencies of parse sdk to 5.1.3


funjay:
### New Issue Checklist

- [x ] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x ] I am not just asking a [question](https://github.com/parse-community/.github/blob/main/SUPPORT.md).
- [ x] I have searched through [existing issues](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).
- [ x] I can reproduce the issue with the latest version of [Parse Server](https://github.com/parse-community/parse-server/releases) and the [Parse Flutter SDK](https://github.com/parse-community/Parse-SDK-Flutter/releases). 

### Issue Description

Pub get fails 
Because no versions of parse_server_sdk_flutter match >6.0.0 <7.0.0 and parse_server_sdk_flutter 6.0.0 depends on parse_server_sdk ^5.1.2, parse_server_sdk_flutter ^6.0.0 requires parse_server_sdk ^5.1.2.
And because parse_server_sdk >=4.0.0 depends on http ^0.13.5, parse_server_sdk_flutter ^6.0.0 requires http ^0.13.5

Parse_server_sdk is now 5.1.3 can we update the sdk_flutter to depend on the latest?

### Steps to reproduce

Add  parse_server_sdk_flutter: ^6.0.0
and any other package that depends on http: > 1.0.0

### Actual Outcome

Fails with error above

### Expected Outcome

Pub get should succeed without errors

### Environment

Parse Flutter SDK
- SDK version: `6.0.0`
- Flutter version: `3.10.6`
- Dart version: `3.0.6`
- Operating system version: `Fedora`

Server
- Parse Server version: `not required`

funjay:
invalid

mbfakourii:
> invalid

I am looking into this possibility, unfortunately, in my last check, we had many conflict with other packages

mbfakourii:
@funjay 

According to [our support policy](https://github.com/parse-community/Parse-SDK-Flutter/tree/master/packages/dart#compatibility), we currently cannot cancel our support for versions below 3 darts(Until May 2024).

My suggestion is to use [dependency_overrides](https://dart.dev/tools/pub/dependencies#dependency-overrides).
like this
```dart
dependency_overrides:
  http: ^1.1.0
```

PR #967, opened 2023-09-25
refactor: Update dart dependencys

mbfakourii:
## Pull Request

## Issue
Update dart dependencys

Closes: #966

## Approach
n/a

## Tasks

- [ ] Add tests


mbfakourii:
@mtrezza 

Considering this [ci](https://github.com/parse-community/Parse-SDK-Flutter/actions/runs/6304370818/job/17115596456) error, I think it is better to stop [supporting](https://github.com/parse-community/Parse-SDK-Flutter/tree/master/packages/dart#compatibility) Dart below 3

mtrezza:
What does our support policy say?

mbfakourii:
> According to our [support policy](https://github.com/parse-community/Parse-SDK-Flutter/tree/master/packages/dart#compatibility), dart 2.x will be supported until March 2024. What is the issue?

We have a problem to update ```http``` package to version above 1
```dart
Because parse_server_sdk depends on http >=1.0.0 which requires SDK version >=3.0.0 <4.0.0, version solving failed.
```

mtrezza:
I believe this is a special case because it's the first stable release 1.0.0, so the developers just decided to cut Dart 2.x support, which is fair enough. The issue is more on our side that we decided to use an unstable library (version 0.x). Maybe we should have looked for an alternative instead. That's a lesson for us when considering dependencies.

Under normal circumstances I don't think we've seen such a drastic drop for support, Dart 3 has been released just 4 months ago. Such short forced upgrade cycles would surely be challenging for developers, so I think our support policy is still good.

I think our options are:
- a) Temporarily look for an alternative to the http library and move to version 1.0.0 in May 2024 when we [end support for Dart 2.19](https://github.com/parse-community/Parse-SDK-Flutter/issues/968).
- b) Keep as is and maybe recommend developers to override the dependency in the meantime on their own risk.
- c) Introduce a new release branch so we have a separate SDK for Dart 2.x and 3.x. --> likely too much work.

mbfakourii:
> I believe this is a special case because it's the first stable release 1.0.0, so the developers just decided to cut Dart 2.x support, which is fair enough. The issue is more on our side that we decided to use an unstable library (version 0.x). This library could have introduced breaking changes with any release, so we took some risk there. We should have looked for an alternative instead. That's a lesson for us when considering dependencies.
> 
> Under normal circumstances I don't think we've seen such a drastic drop for support, Dart 3 has been released just 4 months ago. Such a short forced upgrade cycle is unusual (they are more like years in most ecosystems) and would likely be challenging for developers, so I think our policy for 1 year support is still good.
> 
> I think our options are:
> 
> * a) Temporarily look for an alternative to the http library and maybe move to version 1.0.0 in May 2024 when we [end support for Dart 2.19](https://github.com/parse-community/Parse-SDK-Flutter/issues/968).
> * b) Keep as is and maybe recommend developers to override the dependency in the meantime on their own risk.
> * c) Introduce a new release branch so we have a separate SDK for Dart 2.x and 3.x. --> likely too much work.

I think option b is better. 

There's a method called [`dependency_overrides`](https://dart.dev/tools/pub/dependencies#dependency-overrides) that allows us to temporarily override all references to a dependency.

Regarding the `http` library, I believe it was the first library to handle server requests in Dart, and we were forced to use it at that time. Currently, we not only use `http` but also support `Dio`.

mtrezza:
> Currently, we not only use http but also support Dio.

Can we remove it then?

mbfakourii:
> > Currently, we not only use http but also support Dio.
> 
> Can we remove it then?

No, the ```http``` package is one of the main foundations of our package

mbfakourii:
One of the new features of Dart 3 is [Multiple returns](https://dart.dev/language/records#multiple-returns), which allows us to have multiple output values in the function

Another point is that Flutter and Dart do not have a long lifespan and are developing and growing, I think they offer a new version almost every two weeks.

mtrezza:
Apologies, I've misread your comment.

mbfakourii:
> Apologies, I've misread your comment.

I'm sorry for my language. English is not my native language, but I am improving it. Please let me know if there is anything wrong. ❤️

Regarding this PR, I think it is better to wait for the result of [Dart support policy](https://github.com/parse-community/Parse-SDK-Flutter/issues/968).

mbfakourii:
I think there is no need for this PR anymore

PR #1113, opened 2026-02-25
feat: Expose `ParseDioClient.additionalHeaders` to set additional custom headers applied to all requests

chadpav:
## New Feature

`ParseHTTPClient` already exposes `additionalHeaders` via a public getter/setter, allowing developers to inject custom headers into every request. However, `ParseDioClient` does not expose this, even though the internal `_ParseDioClient` class fully supports it — it reads `additionalHeaders` in its `request()` override and merges them into every outgoing request.

This adds the same 2-line getter/setter forwarding to `ParseDioClient`, bringing it to parity with `ParseHTTPClient`.

### Motivation

Developers who need `ParseDioClient` (e.g., to work around TLS/HandshakeException issues on Android) currently cannot set custom headers without modifying the package source. This small change enables use cases like adding diagnostic or telemetry headers to all Parse requests.

### Changes

- Added `additionalHeaders` getter/setter to `ParseDioClient` that forwards to the internal `_ParseDioClient.additionalHeaders` field

### Example

```dart
final client = ParseDioClient(sendSessionId: true);
client.additionalHeaders = {
  'X-Custom-Header': 'value',
};
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Exposed client-level custom headers on the HTTP client so users can set headers applied to all requests; headers can be set, read back, and cleared.
* **Tests**
  * Added unit tests validating default state, setting/retrieving headers (including nested access), and clearing behavior.

### DIFF
--- a/packages/dart/lib/src/network/parse_dio_client.dart
+++ b/packages/dart/lib/src/network/parse_dio_client.dart
@@ -32,6 +32,15 @@ class ParseDioClient extends ParseClient {
 
   dio.Dio get client => _client;
 
+  /// Custom headers to include in every request made by this client.
+  ///
+  /// This mirrors the same functionality already exposed by [ParseHTTPClient].
+  /// The internal [_ParseDioClient] reads these headers in its [request]
+  /// override and merges them into every outgoing request.
+  Map<String, String>? get additionalHeaders => _client.additionalHeaders;
+  set additionalHeaders(Map<String, String>? additionalHeaders) =>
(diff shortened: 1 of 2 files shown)


mtrezza (commented):
Lint

parseplatformorg:
🎉 This change has been released in version [dart-v9.7.0](https://github.com/parse-community/Parse-SDK-Flutter/releases/tag/dart-9.7.0)

PR #866, opened 2023-04-14
fix: Incorrect results when `ParseQuery` contains special characters

Nidal-Bakir:
### New Pull Request Checklist

- [x] I am not disclosing a [vulnerability](https://github.com/parse-community/Parse-SDK-Flutter/security/policy).
- [x] I am creating this PR in reference to an [issue](https://github.com/parse-community/Parse-SDK-Flutter/issues?q=is%3Aissue).

### Issue Description

Closes: #841

### Approach
use `Uri.encodeComponent(URLComponent);`

This will produce an encoded query value:
`?where={"Name":"AB%2BAB"}`
The encoded equivalent of:
`?where={"Name": "AB+AB"}`

### TODOs before merging

- [x] Add tests
- [x] Add changes to documentation (guides, repository pages, in-code descriptions)
- [x] A changelog entry

PR #1134, opened 2026-04-18
fix: Special characters in list-based `where` constraints break URL parsing

AdrianCurtin:
## Problem

`QueryBuilder` methods that accept a `List<dynamic>` — `whereContainedIn`, `whereNotContainedIn`, and `whereArrayContainsAll` — do not URL-encode String elements before embedding them into the `where={...}` querystring built by `buildQuery()`.

The resulting string is passed to `Uri(query: ...)` in `getSanitisedUri` (`parse_utils.dart`). Dart's `Uri` constructor treats `&`, `=`, `+`, and `#` as valid query sub-delimiters per RFC 3986 and does **not** percent-encode them when they appear inside the `query:` argument. As a consequence, a list element like `"Peanut Butter & Jelly"` ends up on the wire as a literal `&`, causing the server-side querystring parser (Parse Server / Express / `qs`) to split the `where` value at the unescaped delimiter. The JSON becomes malformed and the server rejects the request:

```
ParseError: Invalid parameter for query:  Jelly"]}}
    at ClassesRouter.optionsFromBody
    ...
  code: 102
```

### Minimal reproducer

```dart
final query = QueryBuilder<ParseObject>(ParseObject('Food'))
  ..whereContainedIn('tags', <String>['Peanut Butter & Jelly']);
await query.query(); // => 400 Invalid parameter for query
```

## Fix

Follows the pattern introduced in #866 for the scalar `whereEqualTo` / `whereNotEqualTo` / `whereStartsWith` / `whereEndsWith` / `whereContains` / `whereContainsWholeWord` methods: apply `Uri.encodeComponent` to each String element at constraint-build time. Non-String elements (numbers, pointers, embedded maps, etc.) are passed through unchanged.

A small private helper `_encodeStringElements` is added to avoid duplicating the map-and-encode logic across the three call sites.

## Tests

Two new tests added to `packages/dart/test/src/network/parse_query_test.dart`:

1. **Encoding coverage** — verifies that `&`, `=`, `+`, and `#` are percent-encoded in the `where={...}` string produced by `whereContainedIn`, `whereNotContainedIn`, and `whereArrayContainsAll`, and that the raw unescaped delimiters do **not** appear inside the JSON values.
2. **Non-String pass-through** — verifies that `whereContainedIn('nums', [1, 2, 3])` still produces `"$in":[1,2,3]` with numeric elements intact.

Full `dart test` suite (208 tests) passes locally.

## Related

- #866 — original fix applying `Uri.encodeComponent` to scalar String constraint methods. This PR extends the same approach to the list-based methods that were missed.

## Summary by CodeRabbit

* **Bug Fixes**
  * String values in query builders are now JSON-escaped and percent-encoded so special characters (e.g., &, =, +, #) and escaped quotes/backslashes are transmitted safely, preventing malformed query strings.
  * Non-string list elements (e.g., numbers) are preserved and remain unaffected.

* **Tests**
  * Added tests verifying proper encoding/escaping for string inputs and correct handling of non-string list elements.

### DIFF
--- a/packages/dart/lib/src/network/parse_query.dart
+++ b/packages/dart/lib/src/network/parse_query.dart
@@ -129,7 +129,7 @@ class QueryBuilder<T extends ParseObject> {
     String prefix, {
     bool caseSensitive = false,
   }) {
-    prefix = Uri.encodeComponent(prefix);
+    prefix = _encodeStringElement(prefix);
 
     if (caseSensitive) {
       queries.add(
@@ -155,7 +155,7 @@ class QueryBuilder<T extends ParseObject> {
     String prefix, {
     bool caseSensitive = false,
   }) {
-    prefix = Uri.encodeComponent(prefix);
+    prefix = _encodeStringElement(prefix);
 
     if (caseSensitive) {
       queries.add(
@@ -178,7 +178,7 @@ class QueryBuilder<T extends ParseObject> {
   /// to be equal to the provided [value]
   void whereEqualTo(String column, dynamic value) {
     if (value is String) {
-      value = Uri.encodeComponent(value);
+      value = _encodeStringElement(value);
     }
 
     queries.add(
@@ -237,7 +237,7 @@ class QueryBuilder<T extends ParseObject> {
   /// to be not equal to the provided [value]
   void whereNotEqualTo(String column, dynamic value) {
     if (value is String) {
-      value = Uri.encodeComponent(value);
+      value = _encodeStringElement(value);
     }
 
     queries.add(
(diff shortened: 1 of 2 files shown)


AdrianCurtin:
Worth noting as a security upside of this change: the JSON-escape-then-percent-encode pipeline also closes a latent NoSQL-operator-injection path. Previously, a string value containing `"` was percent-encoded to `%22`, which the server URL-decodes back to a bare `"` inside the `where={...}` JSON literal — an attacker-controlled value could break out of the string and inject operators like `$ne`, `$gt`, `$where`, etc.

With `_encodeStringElement`, `"` is JSON-escaped to `\"` before percent-encoding, so the server sees a well-formed JSON string literal regardless of input. Same goes for `\` and newlines.

Not the primary motivation for the fix, but a nice side effect.

PR #1136, opened 2026-04-28
feat: Multiple SDK-level fixes for anonymous-to-password conversion (iOS parity)

chadpav:
## Issue

No existing issue; problems described below.

## Approach

The Dart SDK's anonymous-to-password user conversion
(`user.username = email; user.password = pw; user.save()` after
`loginAnonymous()`) has three independent gaps vs the iOS SDK. All
three surfaced during empirical testing in a production-mirror setup
as causes of `invalidSessionToken` rejections or orphan `_Session`
row accumulation. This PR ports iOS's behavior into the Dart SDK
across three logically-distinct commits.

### Fix 1 — Reconcile local `authData` on conversion

When a persisted anonymous user has its `username` set and is saved,
`parseUser.save()` issues a `PUT /classes/_User/:objectId` carrying
only the dirty fields. The Dart SDK treated this as a generic
property update — it neither emitted an `authData` unlink signal in
the request body nor cleaned up the local cache after the save.

The iOS Parse SDK handles this via two helpers in `PFUser.m`:

1. **`stripAnonymity`** (`PFUser.m:593-607`): the `username` setter
   writes `authData[anonymous] = NSNull` into the local user. The
   PUT body re-injects the map, so the request carries
   `authData: { anonymous: null }` — Parse Server's documented
   unlink signal.
2. **`cleanUpAuthData`** (`PFUser.m:300-313`): after a successful
   save, the local `authData` has any null entries stripped.

Ported to `ParseUser` as two private helpers:

- `_stripAnonymity()` invoked from the `username` setter. On a
  persisted user it sets `authData['anonymous'] = null` locally; on
  a lazy user (no `objectId`) it removes the entry outright.
- `_cleanUpAuthData()` invoked from `save()` and `update()` between
  the response merge and `_onResponseSuccess()`.

A dartdoc paragraph was added to `loginAnonymous` documenting the
conversion path.

### Fix 2 — Adopt `sessionToken` from the save/update response

Parse Server mints a fresh `_Session` row when `password` is set on
an existing `_User` (with `revokeSessionOnPasswordReset = true`, the
current default in parse-server 9.x). The new `sessionToken` is
embedded in the save response. Without adopting it, the global
session in `ParseCoreData` keeps pointing at the prior session —
which the server just destroyed — and every subsequent request
fails with `invalidSessionToken` until the next login.

This mirrors `PFUser._mergeFromServerWithResult` on iOS, which reads
`PFUserSessionTokenRESTKey` out of the response and installs it.

Implemented as `_adoptResponseSessionTokenIfChanged(String?)`
invoked from `save()` and `update()`. The helper is a no-op when
the response carries no `sessionToken` (e.g. a plain field update),
preserving the existing behavior for non-auth saves.

### Fix 3 — Send `X-Parse-Installation-Id` on all HTTP requests

iOS's `PFURLSessionCommandRunner` attaches the install ID header to
every outgoing request
(`PFCommandURLRequestConstructor.m:_getURLRequestHeadersAsyncForCommand:`).
The Dart SDK previously built the header inline only in the four
user-creation methods (`signUp`, `login`, `loginAnonymous`,
`loginWith`). Every other call — including `PUT` save on an
existing user — went out without it.

Without the install ID on writes, parse-server cannot bind the
`_Session` row it mints during an authData/password change to the
device's installation, and `destroyDuplicatedSessions` skips
cleanup of prior sessions on the same install. Empirically this
leaves one orphan `_Session` row per anonymous-to-password link.

Changes:

- Added `sendInstallationId: bool?` to `ParseNetworkOptions`
  (null/true → attach, false → suppress).
- Hoisted a shared `buildHeaders` helper onto the abstract
  `ParseClient` (`@protected`); both `ParseHTTPClient` and
  `ParseDioClient` inherit and route through it.
- `ParseUser.signUp/login/loginAnonymous/_loginWith` drop their
  inline `_getInstallationId()` and pass
  `sendInstallationId: !doNotSendInstallationID` through options —
  the user-facing opt-out parameter continues to work.

## Tasks

- [x] Add tests

## Behavior change disclosure

1. **Wire — anonymous conversion save**: when `objectId` is set
   and `authData` contains an `anonymous` entry, setting `username`
   and saving now produces a PUT body that includes
   `authData: { anonymous: null }`. Previously no `authData` field
   was sent for that save.
2. **Local cache — anonymous conversion**: setting `username` on
   an anonymous-with-`objectId` `ParseUser` now mutates the local
   `authData` map. After a successful save, the `anonymous` entry
   is removed entirely.
3. **Session token — save/update response**: when a `_User`
   save/update response carries a `sessionToken` field (i.e.
   parse-server minted a new session), the SDK installs it as the
   active session via `ParseCoreData().setSessionId(...)`.
   Previously the new token was merged into the local user object
   but the global session continued pointing at the prior token.
4. **Request headers — all HTTP requests**: every outgoing
   request now carries `X-Parse-Installation-Id` by default.
   Previously only `signUp`, `login`, `loginAnonymous`, and
   `loginWith` set it. Opt-out is preserved via
   `ParseNetworkOptions(sendInstallationId: false)` and the
   existing `doNotSendInstallationID` parameter on user methods.

The commit prefixes mix `fix:` and `feat:`. The PR is labeled
`feat:` overall — the install-ID change is a feature-level default
change, even with opt-out preserved.

If maintainers prefer to split commits `83244367` and `5b2cd09e`
into separate PRs, they are independent of each other and of the
original three commits and can be cherry-picked onto clean branches.

## Documentation follow-up (out of scope for this PR)

This PR adds a dartdoc paragraph to `loginAnonymous` covering the
conversion pattern. The companion user guide at
https://docs.parseplatform.org/dart/guide/ (separate repo) has no
section equivalent to the iOS guide's "Linking Anonymous Users".
Mirroring that section into the Dart guide is a natural follow-up
once this PR lands.

## Tests

Added to `packages/dart/test/src/`:

- `objects/parse_user/parse_user_anonymous_link_test.dart` — 6
  tests covering Fix 1 (authData reconciliation behavior).
- `objects/parse_user/parse_user_session_token_test.dart` (new) —
  3 tests covering Fix 2 (sessionToken adoption from save/update
  response).
- `network/parse_client_test.dart` (new) — 4 unit tests covering
  Fix 3's `buildHeaders` helper (default attach, opt-out via
  `sendInstallationId`, preserve caller-supplied header, merge with
  custom headers).
- `network/parse_dio_client_test.dart` — 1 new integration test
  verifying `buildHeaders` output reaches dio's request pipeline
  via a custom `HttpClientAdapter`.
- `network/parse_http_client_test.dart` — copy-pasted
  "ParseDioClient" group label corrected to "ParseHTTPClient".

Total: 14 new tests across 4 test files. All 221 tests in
`packages/dart/test/` pass.

### DIFF
--- a/packages/dart/lib/src/network/options.dart
+++ b/packages/dart/lib/src/network/options.dart
@@ -1,9 +1,14 @@
 part of '../../parse_server_sdk.dart';
 
 class ParseNetworkOptions {
-  ParseNetworkOptions({this.headers});
+  ParseNetworkOptions({this.headers, this.sendInstallationId});
 
   final Map<String, String>? headers;
+
+  /// When `false`, the client suppresses the `X-Parse-Installation-Id`
+  /// header for this request. `null` (the default) lets the client attach
+  /// the header — matching iOS PFURLSessionCommandRunner behaviour.
+  final bool? sendInstallationId;
   // final ParseNetworkResponseType responseType;
 }
 
--- a/packages/dart/lib/src/network/parse_client.dart
+++ b/packages/dart/lib/src/network/parse_client.dart
@@ -69,6 +69,33 @@ abstract class ParseClient {
 
   @Deprecated("Use ParseCoreData() instead.")
   ParseCoreData get data => ParseCoreData();
+
+  /// Returns `options.headers` with `X-Parse-Installation-Id` attached unless
+  /// the caller opted out via `ParseNetworkOptions.sendInstallationId = false`
+  /// or the header is already set. Installation lookup failures fall through
+  /// silently — a network call should not fail because the install ID could
+  /// not be read.
+  @protected
+  Future<Map<String, String>?> buildHeaders(
+    ParseNetworkOptions? options,
+  ) async {
+    if (options?.sendInstallationId == false) return options?.headers;
+    if (options?.headers?[keyHeaderInstallationId] != null) {
+      return options?.headers;
+    }
+    String? installationId;
+    try {
+      installationId =
+          (await ParseInstallation.currentInstallation()).installationId;
+    } catch (_) {
+      return options?.headers;
+    }
+    if (installationId == null) return options?.headers;
+    return <String, String>{
+      ...?options?.headers,
+      keyHeaderInstallationId: installationId,
+    };
+  }
 }
 
 /// Callback to listen the progress for sending/receiving data.
--- a/packages/dart/lib/src/network/parse_dio_client.dart
+++ b/packages/dart/lib/src/network/parse_dio_client.dart
@@ -47,12 +47,13 @@ class ParseDioClient extends ParseClient {
     ParseNetworkOptions? options,
     ProgressCallback? onReceiveProgress,
   }) async {
+    final Map<String, String>? headers = await buildHeaders(options);
     return executeWithRetry(
       operation: () async {
         try {
           final dio.Response<String> dioResponse = await _client.get<String>(
             path,
-            options: _Options(headers: options?.headers),
+            options: _Options(headers: headers),
           );
 
           return ParseNetworkResponse(
@@ -76,6 +77,7 @@ class ParseDioClient extends ParseClient {
     ProgressCallback? onReceiveProgress,
     dynamic cancelToken,
   }) async {
+    final Map<String, String>? headers = await buildHeaders(options);
     return executeWithRetry(
       operation: () async {
         try {
@@ -85,7 +87,7 @@ class ParseDioClient extends ParseClient {
                 cancelToken: cancelToken,
                 onReceiveProgress: onReceiveProgress,
                 options: _Options(
-                  headers: options?.headers,
+                  headers: headers,
                   responseType: dio.ResponseType.bytes,
                 ),
               );
@@ -116,6 +118,7 @@ class ParseDioClient extends ParseClient {
     String? data,
     ParseNetworkOptions? options,
   }) async {
+    final Map<String, String>? headers = await buildHeaders(options);
     return executeWithRetry(
       isWriteOperation: true,
       operation: () async {
@@ -123,7 +126,7 @@ class ParseDioClient extends ParseClient {
           final dio.Response<String> dioResponse = await _client.put<String>(
             path,
             data: data,
-            options: _Options(headers: options?.headers),
+            options: _Options(headers: headers),
           );
 
           return ParseNetworkResponse(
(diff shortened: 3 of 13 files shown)


chadpav:
@mtrezza There are a few other PR's open. Do we have people merging PR's? You can add me as a maintainer if needed.

chadpav:
@parse-community/ios-sdk-maintenance do we have any active maintainers on this SDK? I can help where possible.

mtrezza:
@chadpav it can take some time until your PR gets reviewed; in the meantime you can test this change out and make sure there are no issues.

chadpav:
> @chadpav it can take some time until your PR gets reviewed; in the meantime you can test this change out and make sure there are no issues.

ok, no problem. I'm running my fork in production. I just wanted to make sure someone was still watching these.

chadpav:
The failing CI build is unrelated to this PR. Do you guys need me to address that?

mtrezza:
@coderabbitai should the issues addressed here be split into separate PRs, to satisfy the rules in https://github.com/parse-community/Parse-SDK-Flutter/pull/1136#issuecomment-4335856554?

chadpav:
Per @mtrezza and CodeRabbit's guidance above, this PR has been split into three independent PRs against `master`:

- **#1138** — `fix: Reconcile local authData on anonymous-to-email conversion` (the original concern of this PR)
- **#1139** — `fix: Adopt sessionToken from save/update response on ParseUser`
- **#1140** — `feat: Send X-Parse-Installation-Id on all HTTP requests`

Each branch is cut from current `upstream/master`, isolates its respective concern, and includes only the tests for that concern. Full test suite passes on all three.

Closing this PR in favor of the split.

mtrezza:
@chadpav Is there any specific order in which these PRs need to or should get merged?

chadpav:
@mtrezza No required order — the three PRs are functionally independent. There are small textual conflicts in `parse_user.dart`, but no semantic dependencies:

- **#1138 ↔ #1139** — both add one helper call inside `save()` / `update()`'s `if (response.success) { … }` block.
- **#1138 ↔ #1140** — both touch `loginAnonymous` (the dartdoc + `_keyAuthAnonymous` constant from #1138 sits adjacent to the `sendInstallationId` wiring in #1140).
- **#1139 ↔ #1140** — clean, no overlap.

Merge in whatever order is convenient. I'll rebase the remaining PRs immediately after each merge so the next one lands conflict-free. If you'd like to minimize rebase touchpoints, **#1139 → #1140 → #1138** keeps #1138 as the last to land, since it's the only one that conflicts with either of the others.

PR #1139, opened 2026-05-17
fix: Adopt sessionToken from save/update response on ParseUser

chadpav:
## Issue

No existing issue.

## Approach

Split out from #1136 at maintainer request. This PR isolates the
session-token adoption concern.

Parse Server mints a fresh `_Session` row when `password` is set on
an existing `_User` (with `revokeSessionOnPasswordReset = true`, the
current default in parse-server 9.x). The new `sessionToken` is
embedded in the save response. Without adopting it, the global
session in `ParseCoreData` keeps pointing at the prior session —
which the server just destroyed — and every subsequent request
fails with `invalidSessionToken` until the next login.

This mirrors `PFUser._mergeFromServerWithResult` on iOS, which reads
`PFUserSessionTokenRESTKey` out of the response and installs it.

Implemented as `_adoptResponseSessionTokenIfChanged(String?)`
invoked from `save()` and `update()`. The helper is a no-op when
the response carries no `sessionToken` (e.g. a plain field update
with no password change), preserving the existing behavior for
non-auth saves.

## Tasks

- [x] Add tests

## Behavior change disclosure

When a `_User` save/update response carries a `sessionToken` field
(i.e. parse-server minted a new session), the SDK installs it as the
active session via `ParseCoreData().setSessionId(...)`. Previously
the new token was merged into the local user object but the global
session continued pointing at the prior token.

## Tests

`packages/dart/test/src/objects/parse_user/parse_user_session_token_test.dart`
— 3 tests covering:

- `save()` installs a new `sessionToken` from the response into the global session
- `save()` is a no-op when the response carries no `sessionToken`
- `update()` installs a new `sessionToken` from the response

All 209 tests in `packages/dart/test/` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Global session tokens are now correctly updated from successful server responses during user save and update operations, preventing stale session state and ensuring subsequent requests use the latest token.
  * Responses without a new session token no longer overwrite the existing session state.

* **Tests**
  * Added coverage for session token updates during save and update operations, including responses with changed, unchanged, and omitted tokens.

### DIFF
--- a/packages/dart/lib/src/objects/parse_user.dart
+++ b/packages/dart/lib/src/objects/parse_user.dart
@@ -493,8 +493,10 @@ class ParseUser extends ParseObject implements ParseCloneable {
     if (objectId == null) {
       return await signUp();
     } else {
+      final String? tokenBefore = sessionToken;
       final ParseResponse response = await super.save();
       if (response.success) {
+        _adoptResponseSessionTokenIfChanged(tokenBefore);
         await _onResponseSuccess();
       }
       return response;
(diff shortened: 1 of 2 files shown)


chadpav:
@mtrezza anything I can do to help get these across? This sessionToken not getting saved is a high severity issue in the Flutter SDK.

PR #1140, opened 2026-05-17
feat: Send X-Parse-Installation-Id on all HTTP requests

chadpav:
## Issue

No existing issue.

## Approach

Split out from #1136 at maintainer request. This PR isolates the
network-layer concern.

iOS's `PFURLSessionCommandRunner` attaches `X-Parse-Installation-Id`
to every outgoing request
(`PFCommandURLRequestConstructor.m:_getURLRequestHeadersAsyncForCommand:`).
The Dart SDK previously built the header inline only in the four
user-creation methods (`signUp`, `login`, `loginAnonymous`,
`loginWith`). Every other call — including `PUT` save on an
existing user — went out without it.

Without the install ID on writes, parse-server cannot bind the
`_Session` row it mints during an authData/password change to the
device's installation, and `destroyDuplicatedSessions` skips
cleanup of prior sessions on the same install. Empirically this
leaves one orphan `_Session` row per anonymous-to-password link.

Changes:

- Extended use of the existing `sendInstallationId: bool?` field on `ParseNetworkOptions`
  (null/true → attach, false → suppress) to the new `buildHeaders` code paths —
  this field was already present before this PR.
- Hoisted a shared `buildHeaders` helper onto the abstract
  `ParseClient` (`@protected`); both `ParseHTTPClient` and
  `ParseDioClient` inherit and route through it.
- `ParseUser.signUp/login/loginAnonymous/_loginWith` drop their
  inline `_getInstallationId()` and pass
  `sendInstallationId: !doNotSendInstallationID` through options —
  the user-facing opt-out parameter continues to work.

## Tasks

- [x] Add tests

## Behavior change disclosure

Every outgoing HTTP request now carries `X-Parse-Installation-Id`
by default. Previously only `signUp`, `login`, `loginAnonymous`, and
`loginWith` set it. Opt-out is preserved via
`ParseNetworkOptions(sendInstallationId: false)` and the existing
`doNotSendInstallationID` parameter on user methods.

The commit is labeled `feat:` because this is a default behavior
change, even with opt-out preserved.

## Tests

- `packages/dart/test/src/network/parse_client_test.dart` (new) —
  4 unit tests covering `buildHeaders` (default attach, opt-out
  via `sendInstallationId`, preserve caller-supplied header, merge
  with custom headers).
- `packages/dart/test/src/network/parse_dio_client_test.dart` —
  1 new integration test verifying `buildHeaders` output reaches
  dio's request pipeline via a custom `HttpClientAdapter`.
- `packages/dart/test/src/network/parse_http_client_test.dart` —
  copy-pasted "ParseDioClient" group label corrected to
  "ParseHTTPClient".

All 211 tests in `packages/dart/test/` pass.

## Summary by CodeRabbit

* **New Features**
  * Outgoing requests now include the installation ID by default, including login and signup requests.
  * Added an option to suppress the installation ID header for individual requests.
  * Caller-provided installation ID headers are preserved regardless of capitalization.

* **Bug Fixes**
  * Standardized installation ID header handling across supported HTTP clients.

* **Tests**
  * Added coverage for default, suppressed, custom, and cross-client header behavior.

### DIFF
--- a/packages/dart/lib/src/network/options.dart
+++ b/packages/dart/lib/src/network/options.dart
@@ -1,9 +1,14 @@
 part of '../../parse_server_sdk.dart';
 
 class ParseNetworkOptions {
-  ParseNetworkOptions({this.headers});
+  ParseNetworkOptions({this.headers, this.sendInstallationId});
 
   final Map<String, String>? headers;
+
+  /// When `false`, the client suppresses the `X-Parse-Installation-Id`
+  /// header for this request. `null` (the default) lets the client attach
+  /// the header — matching iOS PFURLSessionCommandRunner behaviour.
+  final bool? sendInstallationId;
   // final ParseNetworkResponseType responseType;
 }
 
--- a/packages/dart/lib/src/network/parse_client.dart
+++ b/packages/dart/lib/src/network/parse_client.dart
@@ -69,6 +69,33 @@ abstract class ParseClient {
 
   @Deprecated("Use ParseCoreData() instead.")
   ParseCoreData get data => ParseCoreData();
+
+  /// Returns `options.headers` with `X-Parse-Installation-Id` attached unless
+  /// the caller opted out via `ParseNetworkOptions.sendInstallationId = false`
+  /// or the header is already set. Installation lookup failures fall through
+  /// silently — a network call should not fail because the install ID could
+  /// not be read.
+  @protected
+  Future<Map<String, String>?> buildHeaders(
+    ParseNetworkOptions? options,
+  ) async {
+    if (options?.sendInstallationId == false) return options?.headers;
+    if (options?.headers?[keyHeaderInstallationId] != null) {
+      return options?.headers;
+    }
+    String? installationId;
+    try {
+      installationId =
+          (await ParseInstallation.currentInstallation()).installationId;
+    } catch (_) {
+      return options?.headers;
+    }
+    if (installationId == null) return options?.headers;
+    return <String, String>{
+      ...?options?.headers,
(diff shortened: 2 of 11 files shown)


chadpav:
In your update to the description you stated that the sendInstallationId  was a new addition. This is not correct. That option was already present, I just refactored how it gets applied so the new methods where we added installationId headers respected that setting.

chadpav:
Side note: from doc comments it sounds like suppressing installationId headers was needed by some flutter web users where adding custom headers was prohibited. Not respecting that setting would break those users.

chadpav:
@coderabbitai[bot] did you check my coments?

chadpav:
@coderabbitai[bot] i mostly wanted to make sure you updated the PR description in case this language winds up in the release notes somewhere. This is an important distinction. Who writes the release description in this process?

chadpav:
Heads up — the `Test Dart beta` failure on this PR isn't from changes here.

`dart analyze packages/dart --fatal-infos` on the Dart beta channel (`3.13.0-103.1.beta`) flags two `use_super_parameters` infos against pre-existing upstream code:

- `packages/dart/lib/src/objects/parse_config.dart:5:3`
- `packages/dart/lib/src/objects/parse_function.dart:7:3`

Neither file is in this PR's diff. The stable Dart 3.10 and Flutter 3.38 jobs all pass; only the beta toolchain's newer lint catches it, and `--fatal-infos` promotes the info to a failing check.

Happy to open a separate small PR converting those two constructors to super-parameters so `master` is clean on Dart beta and unrelated PRs stop being blocked by it. Let me know if you'd like me to do that.

PR #1166, opened 2026-08-10
fix: Beta CI lanes fail on analyzer findings and on files rewritten by CI itself

pastordee:
## Pull Request

## Issue

No existing issue — this is CI-only. Happy to open one first if you'd prefer it tracked that way.

`Test Dart beta` and `Test Flutter beta` currently fail on `master` ([`80553169`](https://github.com/parse-community/Parse-SDK-Flutter/commit/80553169)) and consequently on every open PR. Both are required checks, so nothing can go green while this stands. I hit it on #1165 and offered there to send the workflow fix separately.

## Approach

`dart pub publish --dry-run` exits **65 on any validation warning**, and it counts a dirty tracked tree as one. Both test jobs modify tracked files before reaching that step, so on the beta SDKs the dry run fails on changes CI made itself — not on anything wrong with the package.

**`check-dart`** — `Run build_runner` regenerates the checked-in `*.mocks.dart`, and beta's mockito emits different output than stable's:

```
Validating package...
* 3 checked-in files are modified in git.
  Modified files:
  test/src/network/parse_client_retry_integration_test.mocks.dart
  test/src/network/parse_query_test.mocks.dart
  test/src/objects/parse_object/parse_object_test.mocks.dart
Package has 1 warning.
##[error]Process completed with exit code 65.
```

**`check-flutter`** — the beta Flutter SDK rewrites `analysis_options.yaml` and `example/analysis_options.yaml` during `flutter pub get`:

```
* 2 checked-in files are modified in git.
  Modified files:
  analysis_options.yaml
  example/analysis_options.yaml
Package has 1 warning.
##[error]Process completed with exit code 65.
```

Worth stressing: in both lanes the *actual* checks pass. On beta, `dart analyze --fatal-infos` reports `No issues found!` and the failure is purely this validation step.

### Why it can't be fixed by committing different files

The stable lanes pass this same step only because their generated output happens to match what is committed. Committing beta's output would break stable; committing stable's is what we already have and breaks beta. There is no set of committed files that satisfies both, so it has to be fixed in the workflow.

### The change

Restore the committed sources with `git checkout -- .` immediately before the dry run, so the package is validated **as it would be published** rather than as CI left it.

The dry run also moves to after `Run tests`. That keeps the tests running against the freshly generated mocks exactly as they do today — only the publish validation sees the restored tree. Had I left the dry run where it was, the restore would have reverted the mocks *before* the tests, changing what they run against on beta. Coverage output is untracked, so the restore doesn't touch it, and the coverage steps still find `coverage/lcov.info`.

Net effect per job: one new step, and `Publish dry run` moves two positions later.

## Verification

Reproduced and confirmed locally against `master`, since the failure is a property of the tree state rather than of the SDK channel:

```console
$ printf '\n// simulated regeneration difference\n' >> packages/dart/test/src/network/parse_query_test.mocks.dart
$ cd packages/dart && dart pub publish --dry-run > /dev/null 2>&1; echo "exit=$?"
exit=65                      # reproduces the CI failure

$ git checkout -- .          # the new step
$ cd packages/dart && dart pub publish --dry-run 2>&1 | tail -2
Package has 0 warnings.
exit=0
```

YAML parses, and the resulting step order is:

| `check-dart` | `check-flutter` |
|---|---|
| Install dependencies | Install dependencies |
| Run build_runner | Analyze code |
| Analyze code | Lint |
| Lint | Run tests |
| Run tests | **Restore files rewritten by earlier steps** |
| **Restore files rewritten by earlier steps** | Publish dry run |
| Publish dry run | Convert / Upload coverage |
| Convert / Upload coverage | |

I can't run the beta lanes outside your CI, so the real confirmation is this PR's own checks — if `Test Dart beta` and `Test Flutter beta` come back green here, that is the fix demonstrating itself.

## Alternatives considered

- **Commit the regenerated mocks** — breaks the stable lanes, as above.
- **Move the dry run before `build_runner`** — doesn't help `check-flutter`, where `pub get` dirties the tree before any other step.
- **`--force`** — rejected in combination with `--dry-run`.
- **Drop the dry run on beta only** — loses the validation on that lane rather than fixing it.

## Tasks

- [x] Add tests — n/a, CI-only; verification above
- [x] Add changes to documentation (guides, repository pages, code comments) — the reasoning is in comments beside both new steps

Entirely happy to reshape this if you'd rather solve it another way; the goal is just to get the beta lanes reflecting real failures again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of package publishing workflows by running tests first and ensuring clean, consistent results.
  * Improved handling of unsuccessful connection attempts to prevent unnecessary asynchronous errors.

* **Maintenance**
  * Simplified configuration and cloud function setup while preserving existing behavior.
  * Improved reliability and consistency across package validation processes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,10 +55,19 @@ jobs:
         run: dart analyze packages/dart --fatal-infos
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/dart
-      - name: Publish dry run
-        run: cd packages/dart && dart pub publish --dry-run
       - name: Run tests
         run: (cd packages/dart && dart test --coverage=coverage)
+      # `dart pub publish --dry-run` exits non-zero on ANY warning, and a dirty
+      # tracked tree is one of them. The "Run build_runner" step above rewrites
+      # the checked-in `*.mocks.dart` files, and on the beta SDK mockito emits
+      # output that differs from what is committed — so the dry run fails on
+      # changes CI made itself, not on anything wrong with the package. Restore
+      # the committed sources first, after the tests have already run against
+      # the freshly generated ones.
+      - name: Restore files rewritten by earlier steps
+        run: git checkout -- .
+      - name: Publish dry run
+        run: cd packages/dart && dart pub publish --dry-run
       - name: Convert code coverage
         # Needs to be adapted to collect the coverage at all platforms if platform specific code is added.
         if: ${{ always() && matrix.os == 'ubuntu-latest' }}
@@ -129,10 +138,16 @@ jobs:
         run: flutter analyze packages/flutter --fatal-infos
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/flutter
-      - name: Publish dry run
-        run: cd packages/flutter && dart pub publish --dry-run
       - name: Run tests
         run: (cd packages/flutter && flutter test --coverage)
+      # Same reason as the dart job: the dry run fails on a dirty tracked tree,
+      # and on the beta SDK `flutter pub get` rewrites the checked-in
+      # `analysis_options.yaml` (and the example's) during the install step
+      # above. Validate against the committed sources.
+      - name: Restore files rewritten by earlier steps
+        run: git checkout -- .
+      - name: Publish dry run
+        run: cd packages/flutter && dart pub publish --dry-run
       - name: Convert code coverage
         # Needs to be adapted to collect the coverage at all platforms if platform specific code is added.
         if: ${{ always() && matrix.os == 'ubuntu-latest' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,13 +57,9 @@ jobs:
         run: dart format --output=none --set-exit-if-changed packages/dart
       - name: Run tests
         run: (cd packages/dart && dart test --coverage=coverage)
-      # `dart pub publish --dry-run` exits non-zero on ANY warning, and a dirty
-      # tracked tree is one of them. The "Run build_runner" step above rewrites
-      # the checked-in `*.mocks.dart` files, and on the beta SDK mockito emits
-      # output that differs from what is committed — so the dry run fails on
-      # changes CI made itself, not on anything wrong with the package. Restore
-      # the committed sources first, after the tests have already run against
-      # the freshly generated ones.
+      # The dry run fails on a dirty tree, and build_runner above rewrites the
+      # checked-in `*.mocks.dart`. Validate the committed sources instead,
+      # after the tests have used the generated ones.
       - name: Restore files rewritten by earlier steps
         run: git checkout -- .
       - name: Publish dry run
@@ -140,10 +136,8 @@ jobs:
         run: dart format --output=none --set-exit-if-changed packages/flutter
       - name: Run tests
         run: (cd packages/flutter && flutter test --coverage)
-      # Same reason as the dart job: the dry run fails on a dirty tracked tree,
-      # and on the beta SDK `flutter pub get` rewrites the checked-in
-      # `analysis_options.yaml` (and the example's) during the install step
-      # above. Validate against the committed sources.
+      # Same as the dart job: `flutter pub get` above rewrites the checked-in
+      # `analysis_options.yaml`, and the dry run fails on a dirty tree.
       - name: Restore files rewritten by earlier steps
         run: git checkout -- .
       - name: Publish dry run
--- a/packages/dart/lib/src/network/parse_live_query.dart
+++ b/packages/dart/lib/src/network/parse_live_query.dart
@@ -305,7 +305,10 @@ class LiveQueryClient {
         if (_debug) {
           print('$_printConstLiveQuery: Error when connection client');
         }
-        return Future<void>.value(null);
+        // Plain  rather than a Future: this is an async body, so
+        // returning a Future inside the try block trips
+        // unawaited_return_in_try_block on newer analyzers.
+        return null;
       }
       WebSocketChannel channel = webSocket.createWebSocketChannel();
       _channel = channel;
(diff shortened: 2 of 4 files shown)


pastordee:
### Update: the workflow fix works, but the Dart lane needed a second fix too

The first run here was a useful correction to my own diagnosis, so recording it plainly.

**`Test Flutter beta` went green** — the first time it has passed on this repo. That lane had no other blocker, so it demonstrates the workflow change does what it claims.

**`Test Dart beta` still failed**, but *earlier* than the step this PR touches. It never reached the publish dry run at all:

```
Analyzing dart...
warning - lib/src/network/parse_live_query.dart:308:9 - Returning a 'Future' without 'await' inside a try block. - unawaited_return_in_try_block
   info - lib/src/objects/parse_config.dart:5:3   - ... could be super parameters. - use_super_parameters
   info - lib/src/objects/parse_function.dart:7:3 - ... could be super parameters. - use_super_parameters
3 issues found.
##[error]Process completed with exit code 2.
```

So the Dart beta lane has **two** independent causes, and I had only described the second one here:

1. three `--fatal-infos` findings the beta analyzer reports and stable does not (fails at `Analyze code`);
2. the dirty-tree publish validation (fails at `Publish dry run`).

Fixing (2) alone just moves the Dart failure to (1), which is exactly what happened.

I have added (1) to this PR in 3b1ec159, so it now fixes both beta lanes on its own rather than half of one. All three are pre-existing on `master` and behaviour-preserving.

**Overlap with #1165:** that PR also carries these three lint fixes, because it needed its own beta lane to get past `Analyze code`. The changes are identical, so whichever merges first the other rebases cleanly — no action needed, just flagging it so it isn't a surprise. If you would rather this PR stay strictly CI-only, say the word and I will drop 3b1ec159 and let #1165 carry the lints; the trade-off is that this PR then cannot turn the Dart beta lane green by itself.

Locally on stable after the cherry-pick: `dart analyze --fatal-infos` clean, `dart format` clean.

pastordee:
### Both beta lanes are green

| Check | Before | Now |
|---|---|---|
| **Test Dart beta** | ❌ for months | ✅ |
| **Test Flutter beta** | ❌ for months | ✅ |
| Test Dart 3.10 — Ubuntu / macOS / Windows | ✅ | ✅ |
| Test Flutter 3.38 — Ubuntu / macOS / Windows | ✅ | ✅ |
| Auto-Release Dry Run | ✅ | ✅ |

Both lanes have been failing on `master` for months, so this is the first fully green CI run on this repo in a while. Each cause is now fixed by the commit that matches it:

- 56535a6c — the dirty-tree publish validation (both lanes)
- 3b1ec159 — the three beta-only analyzer findings (Dart lane, which failed before it ever reached the publish step)

### One red left: `codecov/patch`

`0.00% of diff hit (target 44.42%)`.

That is the lint commit: three behaviour-preserving constructor/return-statement edits in `parse_live_query.dart`, `parse_config.dart` and `parse_function.dart`, none of which add testable behaviour. `codecov/project` is fine at `44.43% (+0.01%)`.

I have deliberately not written tests to satisfy it, since tests asserting that a constructor still takes the same parameters would be noise. If patch coverage is enforced here, the options are to waive it for this PR, or to drop 3b1ec159 and let #1165 carry the lints — in which case this PR still fixes `Test Flutter beta`, but `Test Dart beta` stays red until #1165 lands. Your call; happy either way.

.github/workflows/ci.yml
```diff
@@ -55,10 +55,19 @@ jobs:
         run: dart analyze packages/dart --fatal-infos
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/dart
-      - name: Publish dry run
-        run: cd packages/dart && dart pub publish --dry-run
       - name: Run tests
         run: (cd packages/dart && dart test --coverage=coverage)
+      # `dart pub publish --dry-run` exits non-zero on ANY warning, and a dirty
```
mtrezza: Nice find; a more concise code comment please

.github/workflows/ci.yml
```diff
@@ -129,10 +138,16 @@ jobs:
         run: flutter analyze packages/flutter --fatal-infos
       - name: Lint
         run: dart format --output=none --set-exit-if-changed packages/flutter
-      - name: Publish dry run
-        run: cd packages/flutter && dart pub publish --dry-run
       - name: Run tests
         run: (cd packages/flutter && flutter test --coverage)
+      # Same reason as the dart job: the dry run fails on a dirty tracked tree,
```
mtrezza: A more concise code comment please

mtrezza (commented):
This is a "fix" PR, as it changes the source files

.github/workflows/ci.yml
pastordee: Thanks — shortened in 7e5ac15c:

```yaml
# The dry run fails on a dirty tree, and build_runner above rewrites the
# checked-in `*.mocks.dart`. Validate the committed sources instead,
# after the tests have used the generated ones.
```

.github/workflows/ci.yml
pastordee: Shortened in 7e5ac15c:

```yaml
# Same as the dart job: `flutter pub get` above rewrites the checked-in
# `analysis_options.yaml`, and the dry run fails on a dirty tree.
```

pastordee:
Retitled to `fix:` — you're right, `3b1ec159` changes `packages/dart/lib`, so this isn't CI-only.

For the record on how that came about: the PR started as workflow-only, and the first run here showed `Test Flutter beta` going green but `Test Dart beta` still failing — at `Analyze code`, before it ever reached the step this PR touches. That lane has two independent causes, and the three `--fatal-infos` findings had to come along for it to go green at all.

Commit types are left as they are, since they're accurate per-commit: `56535a6c` is genuinely `ci:` (workflow only) and `3b1ec159` is `fix:` (source). If you squash-merge, the `fix:` title now governs.

Note these three lint fixes also exist identically in #1165, which needed them for its own beta lane. Whichever merges first, the other rebases cleanly.

